PR #1, opened 2023-10-13
[SNOW-13, SNOW-15] Add scheduled tasks to daily ingest parquet data 

thomasyu888:
- Daily parquet data ingested from external stage on a daily basis
- Hash filenames


elt/synapse_gold.sql
```diff
@@ -153,3 +153,7 @@ CREATE VIEW IF NOT EXISTS synapse_data_warehouse.synapse.user_certified AS
   select ID, USER_NAME, FIRST_NAME, LAST_NAME, EMAIL, LOCATION, COMPANY, POSITION, PASSED
   from user_cert_joined
 ;
+
+// zero copy clone of processed access records
```
BryanFauble, line 157: Interesting that snowflake uses `//` for comments compared to the traditional `--` or `/* */` structure.

elt/synapse_elt.sql
```diff
@@ -1,49 +1,431 @@
...
-)
-PATTERN = '.*userprofilesnapshots/snapshot_date=.*/.*';
+ALTER STAGE IF EXISTS SYNAPSE_PROD_WAREHOUSE_S3_STAGE REFRESH;
+ALTER TASK refresh_synapse_prod_stage_task RESUME;
+
+CREATE TASK IF NOT EXISTS userprofilesnapshot_task
+    SCHEDULE = 'USING CRON 0 0 * * * America/Los_Angeles'
+    USER_TASK_MANAGED_INITIAL_WAREHOUSE_SIZE = 'SMALL'
+AS
+copy into
```
BryanFauble, line 16: nit: Consistency through the file is telling me most of the SQL commands are CAPS

elt/synapse_elt.sql
```diff
@@ -1,49 +1,431 @@
 use role accountadmin;
+USE DATABASE SYNAPSE_DATA_WAREHOUSE;
+USE SCHEMA SYNAPSE_RAW;
 
-CREATE OR REPLACE TASK refresh_synapse_stage_task
+CREATE TASK IF NOT EXISTS refresh_synapse_prod_stage_task
```
BryanFauble, line 5: Would there be an issue the next time we want to modify the `refresh_synapse_prod_stage_task` if we do not have the `OR REPLACE` keyword?

The concern is that `CREATE TASK IF NOT EXISTS refresh_synapse_prod_stage_task` would tell me that this would only create the task the first time, but the only time we could 'update' it is if we delete the task and then run this sql script.

elt/synapse_elt.sql
```diff
@@ -1,49 +1,431 @@
...
+     $1:passed_on as passed_on,
+     $1:stack as stack,
+     $1:instance as instance,
+     NULLIF(
+       regexp_replace (
+       METADATA$FILENAME,
+       '.*certifiedquizrecords\/record_date\=(.*)\/.*',
+       '\\1'),
+       '__HIVE_DEFAULT_PARTITION__'
+     )                         as record_date
```
BryanFauble, line 106: nit: Formatting for this is several tabs over compared to L38 which doesn't have the tabbing

elt/synapse_elt.sql
```diff
@@ -1,49 +1,431 @@
...
+    $1:modified_by as modified_by,
+    $1:version_number as version_number,
+    $1:file_handle_id as file_handle_id,
+    $1:name as name,
+    $1:is_public as is_public,
+    $1:is_controlled as is_controlled,
+    $1:is_restricted as is_restricted,
+    NULLIF(
+      regexp_replace (
+      METADATA$FILENAME,
```
BryanFauble, line 75: nit: L33 has an extra tab for this enclosed parenthesis - this line does not

elt/synapse_elt.sql
```diff
@@ -1,49 +1,431 @@
...
+      '.*nodesnapshots\/snapshot_date\=(.*)\/.*',
+      '\\1'),
+      '__HIVE_DEFAULT_PARTITION__'
+    )                         as snapshot_date
+  from @synapse_prod_warehouse_s3_stage/nodesnapshots/)
+pattern='.*nodesnapshots/snapshot_date=.*/.*'
+;
```
BryanFauble, line 82: nit: Push `;` to the line above like L42

elt/synapse_gold.sql
thomasyu888: Ya, it supports all of those, but Im going to standardize to `//`

elt/synapse_elt.sql
thomasyu888: Thats a great point... I still haven't figured out how best to handle this via CI/CD.  We don't necessarily want to replace the task and actually use since snowflake has "time-travelling" abilities and keeps a history of these tasks.

```ALTER TASK```

command.  I think that terraform would be better to use here to create, update and delete resources, so we can leverage the terraform API and state, but the terraform snowflake plugin is still very green.

Currently, I leverage the vscode plugin and just execute the commands.

BryanFauble: So in the past I've used `liquibase` to handle DB schema changes via change-sets. It might be too heavy-handed to use here for a relatively simple use-case. 

The idea behind the tool is that changes would follow this flow:

1. A change set is created to create a new 'thing' in the SQL database
2. The change set is ran through all the environments and get's marked as 'completed' in each environment (Liquibase does the marking)
3. A new change is needed to modify the 'thing' in the SQL database so a new change set is created to make those schema changes.

The idea is that it gave us a predictable way to re-create the exact schema of a DB across envs.
--

That being said - for this use case, if we want to do the `ALTER TASK` approach from local to sync these changes up please make sure that the steps & required access are documented somewhere.

thomasyu888: Agreed, I think there are also tools like `dbt` to potentially help with that. 

Eventually, I see us shifting to an actual workflow engine to track changes and scheduling these queries instead of using snowflake queries.  There is a limitation with snowflake tasks and perhaps Airflow would assist. 

That said, all too heavy-handed currently as we are still in a PoC phase.

BryanFauble (approved):
Changes LGTM!

PR #14, opened 2024-01-10
[SNOW-34] Add in file inventory data

thomasyu888:
**Problem:**

File inventory snapshots from AWS needs to be ingested into snowflake

**Solution:**

Create initial ingestion into snowflake


BryanFauble (approved):
LGTM!

PR #15, opened 2024-01-10
[SNOW-42] Append to file inventory task

thomasyu888:
**Problem**

The file inventory table needs to be continually appended to

**Solution**

Add a snowflake task to continuously read from the external stage to append to the table

Depends on #14


BryanFauble (approved):
LGTM!

PR #16, opened 2024-01-12
[SNOW-43] Fix permissions for data engineer role

thomasyu888:
**Problem**

`ALTER` and `DROP` table actions require ownership of the table.  Most created tables right now are by the SYSADMIN role.

**Solution**

Grant the DPE_service user access to the SYSADMIN role.  Eventually we will want to do an overall permissions scrub or transfer everything in snowflake to the DATA_ENGINEER role to keep things clean.


BryanFauble (approved):
LGTM!

PR #19, opened 2024-01-23
[SNOW-38] Add session id for file download records

thomasyu888:
Add session id column for file download records


BryanFauble (approved):
LGTM!

PR #21, opened 2024-01-25
Update permissions for data engineer role

thomasyu888:
**Problem**

Currently the data engineer role doesn't have enough permission to do work within snowflake

**Solution**
Update permissions.


BryanFauble (approved):
LGTM!

PR #27, opened 2024-02-28
IT-3448: procedure to simplify listing file downloaders

xschildw:
This PR creates a procedure to simplify list who dowloaded files within one or more subtrees.

Example: 'call list_downloaders('2021-01-01', '25982471')

Returns a table:
|USER_ID|USER_NAME|EMAIL|SYNAPSE_PROFILE|NUM_DOWNLOADS|EARLIEST_DOWNLOAD_TIME|LATEST_DOWNLOAD_TIME|
|---------|-------------|------|--------------------|--------------------|-------------------------------|-------------------|
|1223456|myUser|myuser@myDomain.com|https://www.synapse.org/#!Profile:123456|10|2023-01-14 12:01:00|2023-02-01 01:00:04|


synapse_data_warehouse/synapse/procs/R2.13.0__add_list_downloaders.sql
```diff
@@ -0,0 +1,41 @@
+CREATE OR REPLACE PROCEDURE list_downloaders(start_record_date string, entity_list string)
```
thomasyu888, line 1: Could we use the `DATE` type here?  string is also fine, but you can do something like

`record_date > DATE('2024-02-16')`

synapse_data_warehouse/synapse/procs/R2.13.0__add_list_downloaders.sql
```diff
@@ -0,0 +1,41 @@
...
+    select fd.user_id, fd.timestamp, fd.association_object_id as entity_id
+    from filedownload fd
+    join filetree ft on ft.id = fd.association_object_id
+    where fd.association_object_type = ''FileEntity''
+        and ft.node_type = ''file''
+        and fd.record_date >= ''' ||:start_record_date|| '''
+    ),';
+    query_str3 varchar default
+    'download_summary (user_id, c, min_t, max_t) as (
+    select user_id, count(*) as c, min(timestamp) as min_t, max(timestamp) as max_t
```
thomasyu888, line 28: The number of downloads here isn't de-duplicated.  A download is considered as 

```
select distinct user_id, record_date, file_handle_id from synapse_data_warehouse.synapse.filedownload
```

synapse_data_warehouse/synapse/procs/R2.13.0__add_list_downloaders.sql
```diff
@@ -0,0 +1,41 @@
...
+    where id in (' || :entity_list || ')
+    union all
+    select fc.parent_id, fc.id, fc.node_type
+    from node_latest fc
+    join filetree ft on ft.id = fc.parent_id   
+    ),');
+    query_str2 varchar default    
+    'download (user_id, timestamp, entity_id) as (
+    select fd.user_id, fd.timestamp, fd.association_object_id as entity_id
+    from filedownload fd
```
thomasyu888, line 20: So to use this function, I assume you have to do:

```
USE SCHEMA synapse_data_warehouse.synapse # or ...dev.synapse
```

Is that right?

synapse_data_warehouse/synapse/procs/R2.13.0__add_list_downloaders.sql
BryanFauble: +1

synapse_data_warehouse/synapse/procs/R2.13.0__add_list_downloaders.sql
```diff
@@ -0,0 +1,41 @@
...
+    join filetree ft on ft.id = fd.association_object_id
```
BryanFauble, line 21: `node_latest` only contains the entities that have a Synapse ID, however, files stored within a table that have a column type of `File` don't have Synapse IDs. These only have the ID of the file handle, and that file handle ID is attached to the table through a `filehandleassociation` table [(See this related Jira ticket)](https://sagebionetworks.jira.com/browse/PLFM-8336)

Suppose I wanted to get download statistics for one or more of these kinds of files, should this be supported in this stored procedure?

Phrased another way: 
Given a table in Synapse that contains a column of type `File`, suppose I want to retrieve the stats on the downloads for one or more rows in that table. Is there (Or should there be) a stored procedure for this functionality?

BryanFauble (changes requested):
Take a look over the comments when you get a chance. Thanks for the changes Xa

philerooski:
I see a couple of stored procedures which already exist in SYNAPSE_DATA_WAREHOUSE.SYNAPSE that look similar to what we're trying to merge here. Were those procedures created "off the books" (there's no record of their creation in schemachange) and this PR is meant to put one of those procedures back on the books?

image

thomasyu888:
> Were those procedures created "off the books" (there's no record of their creation in schemachange) and this PR is meant to put one of those procedures back on the books?

I think that is what happened.  @xschildw created these with his SYSADMIN account directly in the warehouse and then backfilled this PR.  This is the first stored procedure to be added through schemachange.  @philerooski i wonder if we can create a Jira ticket to finalize this.

PR #31, opened 2024-03-28
[SYNSD-893] Add query for disjointed name

BryanFauble:
**Problem:**
- Synapse allows an entity name to be disjointed from the name of the file. This can lead to some problems when the downloadAs name (The name of the file) overlaps with other files in the folder - and when the name of the FileEntity appears to be correct.

**Solution:**
- Adding in a small query I used to identify which Files in a specific folder had a DownloadAs name which did not match the name of the entity. We had used this to pass along information over to the collaborator to resolve the issue.

I also added a small script via this commit to the examples section that contains how to update all files in a folder to match the downloadAs to the name of the entity: https://github.com/Sage-Bionetworks/synapsePythonClient/commit/ed00ef506e5d5334c21728e4533e57a52124bd3e


BryanFauble:
@thomasyu888 Is this the best location to put these 1-off scripts. Is there something else you had in mind, or perhaps if we should create another folder to contain these?

PR #32, opened 2024-04-02
[IBCDPE-832] Add script to get data size for Synapse entity

danlu1:
This PR includes a sql script that can print folder structure for Synapse folders and get the total data size in GiB for each folder.


analytics/folder_stats.sql
```diff
@@ -0,0 +1,141 @@
...
+FOLDER_IDs can be a single Synapse folder ID or a list of Synapse folder IDs seperated by comma
+*/
+
+USE ROLE DATA_ANALYTICS;
+USE DATABASE synapse_data_warehouse;
+USE WAREHOUSE COMPUTE_XSMALL;
+
+-- The list of folders to be checked
+SET
+    FOLDER_IDs = '';
```
BryanFauble, line 13: This is not working for me. I tried a folder with ID `syn55226438`. In this field, I tried:
`syn55226438` and `55226438`.

In both cases I am getting this error back:
```
error line 0 at position -1
'SYS_VW.PARENT_ID_5' in select clause is neither an aggregate nor in the group by clause.
```

It seems like it is pointing to this statement:
```
SELECT
    'syn' || nodesnapshots.FOLDER_ID AS FOLDER_ID,
    'syn' || nodesnapshots.PARENT_ID AS PARENT_ID,
    'syn' || nodesnapshots.ID AS ID,
    COUNT(nodesnapshots.ID) AS COUNTS,
    nodesnapshots.NAME,
    nodesnapshots.NODE_TYPE,
    nodesnapshots.FILE_HANDLE_ID,
    filesnapshots.CONTENT_SIZE
FROM
    nodesnapshots
JOIN 
    synapse_data_warehouse.synapse.file_latest AS filesnapshots 
ON nodesnapshots.file_handle_id = filesnapshots.ID
GROUP BY nodesnapshots.FOLDER_ID;
```

BryanFauble:
I verified that the Folder content size summing worked as expected. I was happy to see that I could input a Project and sum an entire project size as well.
I used https://www.synapse.org/#!Synapse:syn49637038/files/ as an example. I verified against a query I wrote that the numbers match (22393.32478058 GiB = 21.868481231035155 TiB)
image

image

analytics/folder_stats.sql
```diff
@@ -0,0 +1,141 @@
+/* 
+This script contains two parts: 1. lay out folder structure of the targeted FOLDER_IDs;
```
BryanFauble, line 2: This SQL script contains these 2 parts 1/2 - I would separate them into different .sql files since they serve different purposes.

@thomasyu888 [Also posed the question to me on a SQL script I wrote](https://github.com/Sage-Bionetworks/snowflake/pull/31/files#r1545533636):
> This is great! What are your thoughts on creating a stored procedure? Similar to #27

I am going to be looking at creating a stored procedure for my script - And convert my `.sql` file into an example of using it.

thomasyu888:
This is fantastic!  Same comment as the other PR, but this might be a helpful stored procedure so people can use this feature with ease!  Maybe mimic #27

analytics/folder_stats.sql
danlu1: Thanks a lot for catching this. I have to look at it in depth. I don't remember seeing this error before. Maybe I copied the wrong script.

BryanFauble (approved):
LGTM!

PR #33, opened 2024-04-11
create AD file annotation view

avanlinden:


analytics/ad_node_annotations.sql
```diff
@@ -0,0 +1,52 @@
...
+    ANNOTATIONS:annotations:modelSystemType:value[0] AS MODELSYSTEMTYPE,
+    ANNOTATIONS:annotations:libraryPrep:value[0] AS LIBRARYPREP,
+    -- add some annotations that are not in the current fileview
+    ANNOTATIONS:annotations:cohort:value AS COHORT,
+    ANNOTATIONS:annotations:dataContributionGroup:value AS DATACONTRIBUTIONGROUP, -- noqa: LT05
+    ANNOTATIONS:annotations:dataGenerationSite:value[0] AS DATAGENERATIONSITE,
+    ANNOTATIONS:annotations:isSampleExchange:value[0] AS ISSAMPLEEXCHANGE,
+    ANNOTATIONS:annotations:batch:value[0] AS BATCH
+FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
+WHERE PROJECT_ID = '2580853' AND NODE_TYPE = 'file';
```
thomasyu888, line 52: @avanlinden were there folders that had annotations?

avanlinden: There are but we don't use them any more/wouldn't use them for the same things we use file annotations for, and they're a real grab bag -- they don't conform to the fileview schema.

avanlinden:
Figured I would move this branch to our adkp/snowflake fork so it doesn't hang out here stagnating forever. https://github.com/adknowledgeportal/snowflake/tree/ad-annotation-view

PR #34, opened 2024-04-15
Extract ip addresses

thomasyu888:
**Problem**
It would be nice if we gathered the individual IP address information so we can make interesting analysis around cost and usage.

**Solution**
Leveraging a free API to extract the IP information and pushing it to a snowflake table.

Admittedly, the code is a bit hacky right now...

### DIFF
--- a/sage/ip_lookup.py
+++ b/sage/ip_lookup.py
@@ -55,17 +55,15 @@ def main():
 
     query = """
     select
-        distinct processedaccess.x_forwarded_for as unique_ips
+        distinct x_forwarded_for as unique_ips
     from
-        synapse_data_warehouse.synapse.filedownload
-    inner join
         synapse_data_warehouse.synapse.processedaccess
-        on filedownload.session_id = processedaccess.session_id
     where
-        filedownload.record_date between DATE('2024-01-01') and DATE('2024-03-31') and
-        processedaccess.record_date between DATE('2024-01-01') and DATE('2024-03-31');
+        x_forwarded_for is not null and
+        x_forwarded_for not in (select ip from sage.audit.extracted_ip_info) and
+        record_date > DATE('2024-01-01');
+        // record_date BETWEEN DATE('2023-01-01') and DATE('2023-06-30');
     """
-
     cs.execute(query)
     unique_ips = cs.fetch_pandas_all()
     # API only takes a batch size of 100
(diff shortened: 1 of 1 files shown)


sage/ip_lookup.py
```diff
@@ -0,0 +1,94 @@
...
+
+cs.execute(query)
+unique_ips = cs.fetch_pandas_all()
+import time
+
+
+def get_ip_info(ip_list: list):
+    while True:
+        try:
+            ip_info_response = requests.post('http://ip-api.com/batch', json=ip_list)
```
BryanFauble, line 49: It might be overengineering a solution to the problem, but if you wanted to start trying some async stuff here you could split these up and use
https://superfastpython.com/asyncio-gather/
https://www.python-httpx.org/async/

sage/ip_lookup.py
```diff
@@ -0,0 +1,94 @@
...
+        # return ip_info
+
+batch_size = 100
+result = []
+
+
+for batch_number, batch_df in unique_ips.groupby(np.arange(len(unique_ips)) // batch_size):
+    print(batch_number)
+    if batch_number <= 72:
+        continue
```
thomasyu888, line 69:
```suggestion
```

sage/ip_lookup.py
thomasyu888: Thanks @BryanFauble  - I'll explore when I have a chance.  What I'm dealing with here is an API rate limit of 15 per minute (the documentation says 45 per minute, but that's not what it is)

So doing an async would almost be too efficient, but educational

sage/ip_lookup.py
```diff
@@ -0,0 +1,105 @@
...
+    Returns:
+        dict: IP information like city, country, region, lat, long, asn, etc
+    """
+    # HACK this is not great...
+    while True:
+        try:
+            ip_info_response = requests.post('http://ip-api.com/batch', json=ip_list)
+        except Exception as err:
+            print(ip_info_response.status_code)
+            if ip_info_response.status_code == 429:
```
BryanFauble, line 34: Future change would be to use the backoff package.

https://pypi.org/project/backoff/

sage/ip_lookup.py
```diff
@@ -0,0 +1,105 @@
...
+    query = """
+    select
+        distinct x_forwarded_for as unique_ips
+    from
+        synapse_data_warehouse.synapse.processedaccess
+    where
+        x_forwarded_for is not null and
+        x_forwarded_for not in (select ip from sage.audit.extracted_ip_info) and
+        record_date > DATE('2024-01-01');
+        // record_date BETWEEN DATE('2023-01-01') and DATE('2023-06-30');
```
BryanFauble, line 65: It's personal preference but I like the idea of throwing these into constants at the top of the script so it's clear what "levers" were meant to be changeable.

BryanFauble (approved):
Approving script in its current state since I know it's a hack to get the first round of data in the door. If these becomes an airflow DAG the items noted should be addressed.

PR #36, opened 2024-05-06
[SNOW-72, SNOW-67] add revoke fields to certification quiz snapshot / latest

thomasyu888:
**Problem**

There are new columns being added that we want to incorporate from the parquet

**Solution**
Add the columns

**Bonus**
Add in file handle s3 bucket.  This was done in Snowsite to test the permissions

Depends on https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/pull/623


synapse_data_warehouse/synapse_raw/tables/V2.15.0__add_columns_to_cert.sql
```diff
@@ -0,0 +1,4 @@
+USE SCHEMA {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP
```
jaymedina, line 1: Why is `accountadmin` not used when adding these columns?

thomasyu888: Great question - by default, it uses SYSADMIN roll to modify our table objects.

BryanFauble (approved):
LGTM!

PR #37, opened 2024-05-07
[SNOW-69] Add stack env variable and fileassociation tables

thomasyu888:
**Problem**
Missing file association table

**Solution**
Append fileassociation table


synapse_data_warehouse/synapse_raw/tasks/V2.16.2__add_file_association_tasks.sql
```diff
@@ -0,0 +1,21 @@
+use role accountadmin;
+use schema {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP
+alter task refresh_synapse_warehouse_s3_stage_task suspend;
+create task if not exists append_to_fileinventory_task
```
thomasyu888, line 4:
```suggestion
create task if not exists append_to_fileassociation_task
```

PR #40, opened 2024-05-14
[SNOW-28] Add in ACL latest table

thomasyu888:
**Problem**
We need a way to obtain the latest state of ACL for entities

**Solution**
- Use dynamic table to create said table
- parse out the resource access json and flatten it so that it is more usable across entities


BryanFauble (approved):
Nice to see how easy the declaration of a dynamic table is

thomasyu888:
> @thomasyu888 Thanks for working on this. Just out of curiosity, what does the `#` mean in the accessType and principalId?

@danlu1 no problem!  So basically if you query the `aclsnapshots` table, they have access types like
```
[{"accesstype#2":["READ","SEND_MESSAGE"],"principalid":00000},{"accesstype#2":["READ","SEND_MESSAGE"],"principalid":00000},{"accesstype#2":["READ","UPDATE","DELETE","TEAM_MEMBERSHIP_UPDATE","SEND_MESSAGE"],"principalid":0000}]
```

in my dynamic table query, I flatten this out into separate rows and parse out the headers `#accesstype#2` and etc.  Did that help?

PR #42, opened 2024-05-18
(WIP) Add GX data validation for synapse data warehouse

thomasyu888:
**Problem**
There is currently very little data observability into the warehouse data, it is essential we run a validation suite so we peak into data validity within our warehouse

**Solution**
Add GX as our data validation layer.
* Add snowflake data source
* Add expectation suites
* Create validator to execute the validation.
image

**Tech Debt**
Archive tf scripts for now so it doesn't confuse admins


BryanFauble (approved):
This is awesome. I'll go ahead and pre-approve here.

Are we able to run every table through these expectation suites, or is it set up to run them all after merge?

thomasyu888:
> This is awesome. I'll go ahead and pre-approve here.
> 
> Are we able to run every table through these expectation suites, or is it set up to run them all after merge?

The idea is to run them each table through the expectation suite, but this should eventually live in the Orca repo as a data operation.

PR #43, opened 2024-05-29
[SNOW-79] Add comments to the columns of the file_latest table

BryanFauble:
**Problem:**

1. No comments were present on the FILE_LATEST table

**Solution:**

1. Creating a SQL script to add comments to the columns

### DIFF
--- a/synapse_data_warehouse/synapse/tables/V2.19.0__add_table_file_latest_comments.sql
+++ b/synapse_data_warehouse/synapse/tables/V2.19.0__add_table_file_latest_comments.sql
@@ -1,0 +1,24 @@
+USE SCHEMA {{database_name}}.synapse; --noqa: JJ01,PRS,TMP
+
+COMMENT ON COLUMN FILE_LATEST.change_type IS 'The type of change that occurred on the file, e.g., CREATE, UPDATE, DELETE.';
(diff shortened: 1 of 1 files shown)


synapse_data_warehouse/synapse/tables/V2.19.0__add_table_file_latest_comments.sql
```diff
@@ -0,0 +1,24 @@
+USE SCHEMA {{database_name}}.synapse; --noqa: JJ01,PRS,TMP
+
+COMMENT ON COLUMN FILE_LATEST.change_type IS 'The type of change that occurred on the file, e.g., CREATE, UPDATE, DELETE.';
```
thomasyu888, line 3: @BryanFauble Thanks for doing this! Can we add a description for the FILE_LATEST table?

BryanFauble: Can you confirm if this is the right syntax:
https://github.com/Sage-Bionetworks/snowflake/pull/43/commits/c95de2583eabe4d68e2070e31e75e474dc172a54

I got it from https://docs.snowflake.com/en/sql-reference/sql/alter-table - But alas I don't have permissions to test it out.

thomasyu888: Yup, that is the right syntax.  Thanks!

thomasyu888 (approved):
🔥 LGTM!

PR #48, opened 2024-06-05
[SNOW-81] Add comments to certifiedquiz_latest table and columns

philerooski:


synapse_data_warehouse/synapse/tables/V2.19.3__add_certifiedquiz_latest_comments.sql
```diff
@@ -0,0 +1,17 @@
+USE SCHEMA {{database_name}}.synapse; --noqa: JJ01,PRS,TMP
+
+COMMENT ON TABLE CERTIFIEDQUIZ_LATEST IS 'This table contain snapshots of submissions of user verification data by ACT. Snapshots are taken when a submission is created or updated. Note: Snapshots are also taken periodically and independently of the changes. The snapshot_timestamp records when the snapshot was taken.';
```
BryanFauble, line 3: The format for this is slightly different, example:
`ALTER TABLE CERTIFIEDQUIZQUESTION_LATEST SET COMMENT = 'The latest snapshot of the questions of the certification quiz. With each entry representing a question answered by the user during the quiz.';  `

```suggestion
ALTER TABLE  CERTIFIEDQUIZ_LATEST SET COMMENT = 'This table contain snapshots of submissions of user verification data by ACT. Snapshots are taken when a submission is created or updated. Note: Snapshots are also taken periodically and independently of the changes. The snapshot_timestamp records when the snapshot was taken.';
```

philerooski: Am I misunderstanding [the docs](https://docs.snowflake.com/en/sql-reference/sql/comment)?

> Comments can be added to all objects (users, roles, warehouses, databases, tables, etc.).
> `COMMENT [IF EXISTS] ON <object_type> <object_name> IS '<string_literal>';`

The docs mention your method as an alternative:

> In addition to this command, a comment can be added when creating or altering an object:
> * Specify the COMMENT parameter in the [CREATE <object>](https://docs.snowflake.com/en/sql-reference/sql/create) or [ALTER <object>](https://docs.snowflake.com/en/sql-reference/sql/alter) command.

BryanFauble: You're right @philerooski - Your way of doing it looks like one of the possible ways of doing it.

BryanFauble (approved):
LGTM!

synapse_data_warehouse/synapse/tables/V2.19.3__add_certifiedquiz_latest_comments.sql
thomasyu888: @philerooski can you update the table description here? This is the "latest" table, the description you have is for the snapshot table

philerooski: @thomasyu888 The [schema file](https://github.com/Sage-Bionetworks/Synapse-Stack-Builder/blob/develop/src/main/resources/templates/datawarehouse/datawarehouse-config.json) you linked to in the ticket doesn't have a `certifiedquizlatest` schema. Could you link me to the correct schema or describe how this table is different from the snapshot table?

thomasyu888: @philerooski Sorry for the confusion.  This latest table is a de-duplicated version of the snapshot table.  The description that you pulled from the schema is the snapshot tables, which contains snapshots per week and per event.

We write something like this to define the latest tables: https://github.com/Sage-Bionetworks/snowflake/blob/dev/synapse_data_warehouse/synapse/dynamic_tables/V2.17.0__fileassociation_team_latest.sql.

So what @BryanFauble suggested:
"The latest snapshot of the questions of the certification quiz. With each entry representing a question answered by the user during the quiz.'"

philerooski: @thomasyu888 Thanks. I've updated the description. I also modified the comment on column `snapshot_date` because it mentioned about being a partition in the snapshot table, which isn't the case in the latest table.

thomasyu888 (approved):
🔥 LGTM!

PR #49, opened 2024-06-07
Add permission to execute procedure to GOVERNANCE

xschildw:


admin/grants.sql
```diff
@@ -327,6 +327,11 @@ GRANT SELECT ON FUTURE VIEWS IN DATABASE SYNAPSE_DATA_WAREHOUSE
 TO ROLE GOVERNANCE;
 GRANT SELECT ON FUTURE DYNAMIC TABLES IN DATABASE SYNAPSE_DATA_WAREHOUSE
 TO ROLE GOVERNANCE;
+GRANT EXECUTE ON ALL PROCEDURES IN DATABASE SYNAPSE_DATA_WAREHOUSE
```
BryanFauble, line 330: @xschildw Can you link off to the docs that specify what `EXECUTE` is granting here?

What I am finding is that https://docs.snowflake.com/en/user-guide/security-access-control-privileges doesn't define an `EXECUTE` privilege. I am seeing that the examples from https://docs.snowflake.com/en/sql-reference/sql/grant-privilege#roles are showing something like:
`GRANT USAGE ON ALL PROCEDURES IN DATABASE SYNAPSE_DATA_WAREHOUSE TO ROLE GOVERNANCE;`

xschildw: Good catch! I just went with what I remembered from SQL... Will fix.

BryanFauble (approved):
LGTM! Will need testing once deployed.

PR #51, opened 2024-06-11
[SNOW-78] Add evolve schema permission for tables

thomasyu888:
**Problem**
Currently, it's not possible to test all of the changes within the snowsight SQL via a cloned database, because of permissions.

**Solution**
Update permissions


admin/grants.sql
```diff
@@ -227,10 +229,12 @@ TO ROLE MASKING_ADMIN;
 -- Allow for data engineer role to create tasks
 -- https://docs.snowflake.com/en/user-guide/tasks-intro#creating-tasks
 USE ROLE ACCOUNTADMIN;
-GRANT EXECUTE MANAGED TASK ON ACCOUNT TO ROLE DATA_ENGINEER; --noqa: PRS
+GRANT EXECUTE TASK, EXECUTE MANAGED TASK ON ACCOUNT TO ROLE taskadmin;
 GRANT CREATE DATABASE ON ACCOUNT TO ROLE DATA_ENGINEER; --noqa: PRS
+USE ROLE securityadmin;
+GRANT taskadmin TO ROLE DATA_ENGINEER; --noqa: PRS
+GRANT taskadmin TO ROLE SYSADMIN; --noqa: PRS
```
rxu17, line 236: why didn't `sysadmin` need this before? The privileges with `taskadmin`?

admin/grants.sql
```diff
@@ -281,15 +285,16 @@ GRANT
     USAGE
 ON FUTURE SCHEMAS IN DATABASE SYNAPSE_DATA_WAREHOUSE_DEV
 TO ROLE DATA_ENGINEER;
-GRANT INSERT, SELECT, UPDATE ON FUTURE TABLES IN DATABASE SYNAPSE_DATA_WAREHOUSE_DEV
+GRANT INSERT, SELECT, UPDATE, EVOLVE SCHEMA ON FUTURE TABLES IN DATABASE SYNAPSE_DATA_WAREHOUSE_DEV
```
rxu17, line 288: I don't see `MODIFY` or `DELETE` mentioned here. Would we still be able to for example test adding columns to a table in a test database we created and also delete that?

rxu17 (approved):
LGTM! Just a couple of comments

admin/grants.sql
thomasyu888: Good question - I'm just taking this opportunity to also allow SYSADMIN to manage tasks. Prior to this, I would have to assume ACCOUNTADMIN to do so.

BryanFauble (approved):
LGTM!

admin/grants.sql
thomasyu888: `MODIFY` isn't an allowed permission and i'll hold off on `DELETE` and `TRUNCATE`

PR #58, opened 2024-06-13
Adds back accessrequirementsnapshot ingestion

BWMac:
This PR depends on #57.

This adds back the ingestion script and if merrged after #57 this should work because the accessrequirementsnapshot table will already exist.


BryanFauble:
Am I missing something @BWMac Shouldn't I see the file show up in the changes? I see in the first commit you deleted it, but it was added back in your second commit:

image

thomasyu888:
This can actually be closed, because the scheduled task managed to do the first ingestion.

Issue #59, opened 2024-06-14
mixed up column description in userprofile_latest table?


avanlinden:
I was looking at the column comments describing the various table schemas in Snowflake (which are so so helpful to have!) and noticed the `ID` column on the `userprofile_latest` table is maybe inaccurately referring to the certified quiz table instead? I was operating on the assumption that this column was the Synapse user ID. Not urgent, just wanted to check!

https://github.com/Sage-Bionetworks/snowflake/blob/ac56ed26dcc5c4d894e6d29fffd71c56b731711a/synapse_data_warehouse/synapse/tables/V2.19.4__add_userprofile_latest_comments.sql#L9

BryanFauble:
Thanks for the report @avanlinden I made an update and this will be fixed during the next change set run. I'll close this out for now.

avanlinden:
Thanks team! Looks good.

PR #60, opened 2024-06-14
[Issue-59] Correct description of userprofile latest

BryanFauble:
**Problem:**

1. 3 Column descriptions differ from their expected values.
2. https://github.com/Sage-Bionetworks/snowflake/issues/59

**Solution:**

1. Updated the column description for the 3 columns.


jaymedina (approved):
Looks good! Double-checked with https://sagebionetworks.jira.com/wiki/spaces/DW/pages/3034775570/Raw+tables+in+Athena+Snowflake#User-Profile-Snapshot

PR #65, opened 2024-06-19
[IBCDPE-989] Lock numpy and pandas vers in ci.yaml

jaymedina:
This should hopefully fix the failing CI build

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,6 +47,8 @@ jobs:
         shell: bash
         run: |
           pip install schemachange
+          pip install numpy==1.26.4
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -47,6 +47,8 @@ jobs:
         shell: bash
         run: |
           pip install schemachange
```
thomasyu888, line 49: Did we want to lock down a schemachange version as well?

jaymedina: Done!

thomasyu888 (approved):
🔥 LGTM!

PR #66, opened 2024-06-19
[SNOW-33] reintroduce projectsettingsnapshots task

jaymedina:
Re-introducing the `projectsettingsnapshots` task script now that builds are working.

Tested the re-introduction of `LOCATIONS` column with dtype `ARRAY` on a test db (commit: https://github.com/Sage-Bionetworks/snowflake/pull/66/commits/1a860e7a1b22275a4e6de798d17ea877e248a9d8)

image

### DIFF
--- a/synapse_data_warehouse/synapse_raw/tables/R__projectsettingsnapshots_append_task.sql
+++ b/synapse_data_warehouse/synapse_raw/tables/R__projectsettingsnapshots_append_task.sql
@@ -1,0 +1,35 @@
+use role accountadmin;
+use schema {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP
+
+alter task refresh_synapse_warehouse_s3_stage_task suspend;
+create task if not exists append_to_projectsettingsnapshots_task
+    user_task_managed_initial_warehouse_size = 'SMALL'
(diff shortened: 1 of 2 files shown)


synapse_data_warehouse/synapse_raw/tables/R__projectsettingsnapshots_append_task.sql
```diff
@@ -0,0 +1,35 @@
+use role accountadmin;
```
thomasyu888, line 1: Could we try removing this row?

jaymedina: Hmm sure, this is role is going off the template in the Contributions page, though. Wouldn't it give us an error without it? (I shut down my codespace so I can't test this super fast)
image

thomasyu888: @jaymedina . so... the should be updated, but basically, we want to avoid creating anything with the Accountadmin role

thomasyu888 (commented):
Gah, I forgot to hit I approve...

PR #68, opened 2024-07-19
[SNOW-103] Create a `streamlit` app template

jaymedina:
### problem

There is currently no template or guide for generating a Streamlit application the communicates with Snowflake, and launching it via an EC2 instance provisioned through Sage's Service Catalog.

### solution

A new template is introduced with built-in scripts and functions that serve to make the development process easier when creating a Streamlit app, and minimizes overhead when deploying within an EC2 instance.

The structure of the template is as follows:

```
streamlit_template/
├── .streamlit/
│   ├── config.toml
│   └── example_secrets.toml
├── tests/
│   ├── __init__.py
│   └── test_app.py
├── toolkit/
│   ├── __init__.py
│   ├── queries.py
│   ├── utils.py
|   └── widgets.py
├── Dockerfile
├── app.py
├── requirements.txt
└── style.css
```

#### example app using template

image


.gitignore
```diff
@@ -3,3 +3,4 @@
 terraform.tfstate*
 *.csv
 .DS_Store
+streamlit_template/.streamlit/secrets.toml
```
jaymedina, line 6: Have `git` ignore any changes to `secrets.toml` to avoid developers accidentally committing their Snowflake credentials to the repository. Instead we use `example_secrets.toml` that people can copy from to create their own setup.

streamlit_template/toolkit/queries.py
```diff
@@ -0,0 +1,167 @@
...
+SELECT
+    project_id,
+    SUM(content_size) AS total_downloads
+FROM
+    file_content_size
+GROUP BY
+    project_id;
+"""
+
+QUERY_UNIQUE_USERS = """
```
thomasyu888, line 119: This is great - you will see quite a bit of repetition of queries for the different programs/projects

.gitignore
thomasyu888: Nit: I would add py_cache to the .gitignore

thomasyu888 (commented):
Awesome work here! adding my inItial comments.

streamlit_template/tests/sample_data_app.py
```diff
@@ -0,0 +1,202 @@
+import streamlit as st
```
thomasyu888, line 1: Will this be replaced with app.py?

streamlit_template/app.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+# Custom CSS for styling
+with open('style.css') as f:
+    st.markdown(f'<style>{f.read()}</style>', unsafe_allow_html=True)
+
+def main():
+
+    # 1. Retrieve the data using your queries in queries.py
+    entity_distribution_df = get_data_from_snowflake(QUERY_ENTITY_DISTRIBUTION)
```
thomasyu888, line 17: I like and dislike that the app doesn't load until all the queries are done. Is there a giant spinning wheel we can wrap it in in the future?

jaymedina: By default when I launch my example app I see this feature on the top right and the widgets load one at a time  (depending on which query finishes, and what data loads up first, I assume). I can launch the app after Sage Week so we can experiment with this.

image

BryanFauble (approved):
This is really awesome to see!


@jaymedina 

To gauge interest and to get your thought. Do you have interest in getting some kubernetes experience later on by Porting this over to run in the k8s cluster?

https://docs.docker.com/guides/deployment-orchestration/kube-deploy/

My goal is to make deployment easy so we can throw apps like this out there and expose it out to Sage staff on the vpn

streamlit_template/app.py
```diff
@@ -0,0 +1,49 @@
+import numpy as np
+import streamlit as st
+
+from toolkit.queries import (QUERY_ENTITY_DISTRIBUTION, QUERY_PROJECT_DOWNLOADS,
```
BryanFauble, line 4: I would recommend adding a pre commit lint for black and sort (Examples in the SynPy)

streamlit_template/toolkit/queries.py
```diff
@@ -0,0 +1,166 @@
+QUERY_ENTITY_DISTRIBUTION = """
+with htan_projects as (
+    // select distinct cast(replace(NF.projectid, 'syn', '') as INTEGER) as project_id from sage.portal_raw.HTAN
+    select
+        cast(scopes.value as integer) as project_id
+    from
+        synapse_data_warehouse.synapse.node_latest,
+        lateral flatten(input => node_latest.scope_ids) scopes
+    where
+        id = 20446927
```
BryanFauble, line 10: It might make sense to extract this out to an environment variable so if we ever did need to change this ID we wouldn't need to create a new docker image to do it

jaymedina:
> To gauge interest and to get your thought. Do you have interest in getting some kubernetes experience later on by Porting this over to run in the k8s cluster?

Absolutely, that sounds like a great idea! This type of app isn't very resource heavy and it'd be a good way to start developing some guidelines for deployment on k8s.

streamlit_template/README.md
```diff
@@ -114,7 +112,7 @@ as you see fit.
 ### 6. Dockerize your Application
 
 - Update the `requirements.txt` file with the packages used in any of the scripts above.
-- Ensure you have pushed all your changes to your fork of the repository that you are working in.
+- Ensure you have pushed all your changes to your fork of the repository that you are working in (remember not to commit your `secrets.toml` file).
```
BryanFauble, line 115: You could use a `.gitignore` file and add `secrets.toml` to it

streamlit_template/toolkit/queries.py
BryanFauble: With the latest set of changes this was not extracted out to an environment variable. It was moved out to a global variable. What it means is that since this is a containerized application we would need to create a new image to point this at a different ID.

It isn't too big of a problem since it isn't likely to change, however, it's something to be aware of if we need to update the ID we will need to build a new image, pull, and then deploy the new image wherever this is hosted.

BryanFauble (approved):

streamlit_template/README.md
jaymedina: I added the `secrets.toml` to the `.gitignore` at the [base of the repo](https://github.com/Sage-Bionetworks/snowflake/blob/SNOW-103-streamlit-template/.gitignore).

But this brings up a good point: I should probably make a separate `.gitignore` to live under `streamlit_template` since we're treating this like an isolated module and we want all the required components to live in the Docker image.

BryanFauble: Yeah that makes sense. I see folks on stack overflow mentioning this pattern

streamlit_template/toolkit/queries.py
jaymedina: Woops, my apologies for misinterpreting. For clarity: The queries that currently live in `queries.py` are actually just placeholders, and what I had in mind was that people using this template to make their own Streamlit apps would make their own queries from scratch, and use the existing examples as guides for how to construct their queries at the most.

In the future it might be a good idea to create a default `BASE_QUERY` that people can use to search for a single project, at which point we can implement a parameter like `project_id` that users can feed and get their subsetted dataframe, but I would leave that for future work.

BryanFauble (approved):

PR #70, opened 2024-08-14
[SNOW-103] Streamlit App Validation Feedback

jaymedina:
## Introduction
This PR implements feedback given as a result of validating the work done in #68.

## Changelog

Notable changes include: 

- [x] **Feature**: Add `launch.json` with configurations for debugging the Streamlit app and running tests against it
- [x] **Feature**: Convert string queries to functions that _return_ string queries, to enable dynamic querying, and therefore interactive Streamlit app widgets
- [x] **Feature**: Change the _Unique Users_ widget to be interactive, and use this as an interactive widget/query example
- [x] **Documentation**: Update `README.md` to reflect the changes above
- [x] **Documentation**: Encourage containerization within EC2 instance rather than working with the GitHub Container Registry (due to the image's size)
- [x] **Documentation**: Add more verbose instructions on building the docker container (section: **Dockerize your Application**)

## Testing & Preview

image


streamlit_template/README.md
```diff
@@ -123,14 +175,59 @@ as you see fit.
...
 - Navigate to your home directory (`cd ~`).
-- Clone your repository in your desired working directory.
-- Create your `secrets.toml` file again. The Docker image of your Streamlit application will not have the `secrets.toml` for security reasons.
-- Build your Docker image from the Dockerfile in the repository
+- Clone your repository in your desired working directory. Example:
+  ```
+  git clone https://github.com/<your-username>/snowflake.git
+  ```
+  Replace `<your-username>` with the user that the forked repository is under.
+- Create your `secrets.toml` file again, and make sure it's located under the `.streamlit/` folder. By default, the instance should already have `vi` available to use as an editor.
```
BryanFauble, line 183: `and make sure it's located under the `.streamlit/` folder`

In this case it doesn't actually need to exist here. It only needs to be available in the current working directory where you run the `docker run` command

streamlit_template/README.md
```diff
@@ -123,14 +175,59 @@ as you see fit.
...
+- Create your `secrets.toml` file again. By default, the instance should already have `vi` available to use as an editor.
+- Build your Docker image, either from the `Dockerfile` in the repository, or by pulling down your image from the GitHub Container Registry.
+- Run your Docker container from the image, and make sure to have your `secrets.toml` (in the current working directory) mounted and the 8501 port specified, like so:
+  ```
+  docker run \
+    -p 8501:8501 \
+    -v $PWD/secrets.toml:/.streamlit/secrets.toml \
     <image name>
   ```
+- Now your Streamlit application should be sharable via the private IP address of the EC2 instance. To find the private IP address, navigate back to the _Events_ tab when viewing your provisioned EC2: Linux Docker product, and scroll down to `EC2InstancePrivateIpAddress`. Let's say your EC2 instance's private IP address is 22.22.22.222. The URL you can share with users to access the Streamlit app would be http://22.22.22.222:8501/. **Remember that this is a private IP address, therefore your Streamlit app can only be viewed by those connected to Sage's internal network, either through ethernet or VPN.**
```
BryanFauble, line 192: > either through ethernet or VPN.

Is that true that all ethernet traffic in the office is routed to the AWS VPN? I obviously cannot test it, but it sounds a little suspect.

BryanFauble (approved):
Thanks for making these changes!

streamlit_template/README.md
jaymedina: Probably worth checking with IT to verify, but for now I'll remove this bit.

PR #74, opened 2024-09-03
[SNOW-86] Fix deduplication logic for `acl_latest`

jaymedina:
## Problem

Duplicates were found in `acl_latest`, with duplicates being defined as repeated entries of rows with a unique combination of `OWNER_ID` (ID of entity), `PRINCIPAL_ID` (ID of permissions inheritor) and `ACCESS_TYPE` (permissions configuration).

## Solution
The original de-duplication logic involved using the `ROW_NUMBER` window function to enumerate each row for every partition of each unique combination of `OWNER_ID` and `CHANGE_TIMESTAMP`. **This allowed for multiple entries for a given `OWNER_ID` to make it through the `ROW_NUM = 1` filter since different timestamps were being captured**. Further, the rows for every partition were sorted by `CHANGE_TIMESTAMP` in ascending order, `ROW_NUM = 1` being the oldest ACL snapshot.
```
...
  ROW_NUMBER() OVER (
              PARTITION BY OWNER_ID, CHANGE_TIMESTAMP
              ORDER BY CHANGE_TIMESTAMP ASC
          ) AS ROW_NUM
...
QUALIFY ROW_NUM = 1
```

The solution was 
1. To have the `acl_expand` CTE be performed on the `aclsnapshots` table first. While this increases compute time it is necessary to unpack the columns of interest in 2. below.
2. To partition based on the definition of a duplicate: In this case, a **duplicate** is a repeated entry of `OWNER_ID`, `ACCESS_TYPE`, and `PRINCIPAL_ID`. Each row should have a _unique_ combination of those 3 columns. To get the latest ACL for a given unique row, we order the partitions by `CHANGE_TIMESTAMP` in descending order, meaning the latest ACL is the first row and makes it through the `QUALIFY ROW_NUM = 1` filter.

```
...
ROW_NUMBER() OVER (
                    PARTITION BY owner_id, access_type, principal_id
                    ORDER BY change_timestamp DESC
                ) AS row_num
```
**BEFORE**
image

**AFTER**
image

## Testing

For our tests, we want to make sure we have no more duplicates (with duplicates being defined as repeated combinations of `OWNER_ID` `ACCESS_TYPE` and `PRINCIPAL_ID`, and that all the **latest** ACLs for a given `OWNER_ID` are accounted for.

### 1. Ensure there are no duplicates

image

### 2. Ensure that all the ACLs are accounted for

image


synapse_data_warehouse/synapse/dynamic_tables/R__populate_acl_latest_table.sql
```diff
@@ -0,0 +1,47 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE DYNAMIC TABLE IF NOT EXISTS ACL_LATEST
```
thomasyu888, line 3: Since `ACL_LATEST` table exists, we would need to do `CREATE OR REPLACE`

thomasyu888 (approved):
🔥 LGTM!  I'm going to pre-approve this, but I'd like @philerooski  to do a review of this prior to merging

philerooski (commented):
Thanks for the clear explanation of the problem and your solution in the description. It made the PR easy to understand.

synapse_data_warehouse/synapse/dynamic_tables/R__populate_acl_latest_table.sql
```diff
@@ -0,0 +1,47 @@
...
+                ) AS access_type,
+                COALESCE(
+                    value:"principalId"::number,
+                    value:"principalid"::number,
+                    value:"principalid#1"::number
+                ) AS principal_id
+            FROM 
+                {{database_name}}.synapse_raw.aclsnapshots,
+                LATERAL FLATTEN(input => parse_json(resource_access), outer => TRUE)
+        ),
```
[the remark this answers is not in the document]
jaymedina: Hi @philerooski  thanks for the feedback! Agreed, it makes sense to work with temp or transient tables rather than CTEs for clarity. I went with using the temp tables because they're session-specific, but added an extra bit of logic to drop the tables after the dynamic table is created, just in case. Changes are in https://github.com/Sage-Bionetworks/snowflake/pull/74/commits/6b2222d6eacc00dd938da3791bf8afafd260fb72

Let me know if this improves the structure and flow of the script & if you have any other suggestions!

synapse_data_warehouse/synapse/dynamic_tables/R__populate_acl_latest_table.sql
```diff
@@ -0,0 +1,57 @@
+-- 1. Initialize the schema that the dynamic table will live under
+USE SCHEMA {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+-- 2. Create the first temporary table which unpacks the resource_access column
+--    and extracts the access type and principal ID as new columns
+CREATE TEMPORARY TABLE acl_expanded_temp AS
```
thomasyu888, line 6: Just a note: the temporary tables need to be within the dynamic table statement, or else they won't be created every single time.

jaymedina: Ah yeah. Since the purpose of using the temporary tables was to simplify the dynamic table logic, I don't think it's helpful to use temporary/transient tables in this case. What I can do to help with code clarity in the dynamic table logic is add more comments to separate the CTEs to provide more structure.

What do you think of going in that direction @philerooski ?

philerooski: Could you explain what you mean, @thomasyu888 ? How can the dynamic table reference the temporary table `dedup_acl_expanded_temp` without that table being created as part of the session?

My intention with factoring out the CTEs was to reduce the amount of nesting in the CREATE TABLE statement. We would never nest functions three-levels deep across 45 lines in _a single statement_ in Python or R, so why do it in SQL if we don't have to?

thomasyu888: @philerooski So the raw tables are updated with data, so you would have to ensure these temporary tables are created every time the dynamic table queries are run.  Temporary tables contain static information, this particular SQL code isn't run on a schedule, but the code defined in the dynamic tables code is.

Separately, CTE's are meant to serve as "functions", so in my opinion, the above is technically proper usage of CTEs.  now I could eventually see the separation of bronze, silver, gold tables, where there is a middle step of (silver tables) containing the first layer, then the gold tables containing the last.

jaymedina: I was going to suggest using `TARGET_LAG` on the temporary tables, but this seems to be a feature unique to dynamic tables. From what I'm reading, it looks like CTEs are the best use-case for this type of thing, since all the logic that runs on a schedule based on `TARGET_LAG` is under the `CREATE OR REPLACE DYNAMIC TABLE...` logic. 

I really prefer the script the way it's structured with the temporary tables, so I'll keep poking around the docs to see if there's an alternative, but for now what do we think of us going forward with the CTEs and structuring them with comments?

thomasyu888:
I just wanted to add that I appreciate the review @philerooski and work @jaymedina - always continue to offer up new perspectives.

synapse_data_warehouse/synapse/dynamic_tables/R__populate_acl_latest_table.sql
philerooski: @thomasyu888 That makes sense. There's a huge list of [limitations](https://docs.snowflake.com/en/user-guide/dynamic-tables-limitations) that come with using dynamic tables, and weirdly enough temporary tables aren't mentioned except indirectly: "You can’t use dynamic SQL (for example, session variables or unbound variables of anonymous blocks) in the dynamic table’s definition."

Probably it's not worth trying to avoid these behemoth dynamic table definitions.

jaymedina:
CTEs have been re-implemented and I've done a final pass of the code and corrected all the small typos / syntax errors I anticipate would trip up schemachange. Once I get a final approval from you both, I'll go ahead and merge. Thanks for the review!

thomasyu888 (approved):
🔥 LGTM!

philerooski (approved):

PR #77, opened 2024-10-20
Create authentication policies

thomasyu888:
## Problem
After adding jumpcloud and supporting multiple IDPs there was an awkward login period for snowflake

## Solution
This avoids that by setting a `user_authentication_policy` on the account level with the opportunity to set specific user authentication policies for those that may want to test the jumpcloud login

Bonus: move most policies into the POLICY_DB so that policies don't get dropped on accident.

Note: these were run manually because there is currently no CI/CD to run auth policies

### DIFF
--- a/admin/policies.sql
+++ b/admin/policies.sql
@@ -14,6 +14,12 @@
 --     ELSE '*****'
 --   END;
 
+USE ROLE SYSADMIN;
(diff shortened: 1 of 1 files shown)


BryanFauble (approved):
LGTM!

PR #78, opened 2024-10-23
[SNOW-90] Introduce CI job to test `schemachange` updates against a clone

jaymedina:
## problem

Rather than have it be a manual process, automating the testing of `schemachange` updates in our branch before merging it into `dev` will speed up the development life cycle. A design document was made to tackle this problem and is available [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3703472153/SNOW-90+CI+Pipeline+for+Schemachange+Testing+in+Sage+s+Snowflake+Development+Environment?focusedCommentId=3706716206).

## solution

image

- [x] Introduce a new branch label, `test_with_clone`, to trigger the automated process of zero-copy cloning `synapse_data_warehouse_dev` and testing new/modified schemachange scripts against it
- [x] Introduce new CI jobs that reflect the design outlined in the [design document](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3703472153/SNOW-90+CI+Pipeline+for+Schemachange+Testing+in+Sage+s+Snowflake+Development+Environment?focusedCommentId=3706716206)
- [x] Update contribution guidelines to explain these new CI jobs and the corresponding PR labels

## testing

- [x] Test that label `create_clone_and_run_schemachange` triggers the CI job to run
- [x] Test that a clone is made of the `dev` db and _not_ the `prod` db
- [x] Test that a clone is named after the branch
- [x] Test that unlabeling and relabeling the branch just replaces the existing clone, and does not make a new one
- [ ] Test that merging the PR triggers `drop_clone` to run
- [ ] Test that merging the PR DOES NOT trigger `create_clone_and_run_schemachange` to run
- [ ] Test that labeling the merged PR with `create_clone_and_run_schemachange` does not trigger the CI job to run
- [ ] Test that back-to-back commits don't cause problems on Snowflake (addresses this [concern](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3703472153/SNOW-90+CI+Pipeline+for+Schemachange+Testing+in+Sage+s+Snowflake+Development+Environment?focusedCommentId=3707928578))


.github/workflows/test_with_clone.yaml
```diff
@@ -0,0 +1,111 @@
...
+            - name: Sanitize Clone Name
+              run: |
+                CLONE_NAME_SANITIZED="${CLONE_NAME//[^a-zA-Z0-9_]/_}"
+                echo "Clone name has been updated! The clone name will be: ${CLONE_NAME_SANITIZED}"
+                echo "SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE=${CLONE_NAME_SANITIZED}" >> $GITHUB_ENV
+
+            - name: Zero-copy clone the database
+              shell: bash
+              run: |
+                ~/bin/snowsql -r $SNOWSQL_CLONE_ROLE -q "CREATE OR REPLACE DATABASE $SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE CLONE $SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE_ORIG;"
```
philerooski, line 61: Some interesting things I learned while reading the docs on cloning:

* A cloned table does not include the load history of the source table. One consequence of this is that data files that were loaded into a source table can be loaded again into its clones.
* Cloning a database or schema does not clone objects of the following types in the database or schema: External tables, Internal (Snowflake) stages
* The "Object references" section describes how objects which reference a fully-qualified name (like `my_database.my_schema.my_table`) also reference the fully-qualified name in their clone. This could potentially be problematic if someone has defined in the production database a view (for example) which references a stage by its fully-qualified name.

CONTRIBUTING.md
```diff
@@ -0,0 +1,104 @@
...
+
+* Creates a zero-copy clone of the database and runs your proposed schema changes against it.
+* Tests your schema changes on a cloned version of the development database, verifying that your updates work correctly without
+affecting the real development database. After the PR is merged, the clone is automatically dropped to free up resources.
+
+> [!IMPORTANT]
+> Your cloned database is a clone of the development database as it exists at the time of cloning. Please be mindful that **there may be
+> other developers working on their own version of the development database, whose changes may not be reflected in your clone**. Keep an
+> eye out for other PRs with the `create_clone_and_run_schemachange` label, and ensure that you are not performing changes on the
+> same tables to avoid conflicts.
```
philerooski, line 75: If we are cloning the dev database, why do we need to be concerned about what's happening in other clones of the dev database? Is this a case where someone else's changes are pulled in before yours and you need to rebase to get an up to date clone?

.github/workflows/test_with_clone.yaml
```diff
@@ -0,0 +1,111 @@
+---
+name: Test Changes with Cloned DB
+
+on:
+    pull_request:
+        types: [ labeled, closed ]
```
philerooski, line 6: Does this trigger if the PR is created with the label already?

CONTRIBUTING.md
```diff
@@ -0,0 +1,104 @@
...
+Once you've made your changes and committed them locally, push your branch to the remote repository:
+
+```
+git push origin snow-123-new-feature
+```
+
+### 5. Create a Draft Pull Request
+
+In order to initiate automated testing you will need to work on a draft pull request (PR) on GitHub. After pushing your commits to
+the remote branch in Step 4, use the GitHub UI to initate a PR and convert it to draft mode.
```
philerooski, line 57: This seems optional. If so, can we make that more explicit?

.github/workflows/test_with_clone.yaml
philerooski: Also -- what if the PR is already labeled, I make some additional commits to my branch, and I want to `CREATE OR REPLACE ... CLONE` again? Do I need to remove the label and add it again?

.github/workflows/test_with_clone.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+permissions:
+    contents: read
+
+jobs:
+    
+    create_clone_and_run_schemachange:
+        runs-on: ubuntu-latest
+        if: contains(github.event.pull_request.labels.*.name, 'create_clone_and_run_schemachange')
```
philerooski, line 15: Won't this job still run if the PR is closed and it contains the 'create_clone_and_run_schemachange' label?

.github/workflows/test_with_clone.yaml
```diff
@@ -0,0 +1,111 @@
...
+                SNOWSQL_DEST=~/bin SNOWSQL_LOGIN_SHELL=~/.profile bash snowsql-1.2.9-linux_x86_64.bash
+
+            - name: Verify SnowSQL installation
+              run: ~/bin/snowsql -v
+              
+            - name: Sanitize Clone Name
+              run: |
+                CLONE_NAME_SANITIZED="${CLONE_NAME//[^a-zA-Z0-9_]/_}"
+                echo "Clone name has been updated! The clone name will be: ${CLONE_NAME_SANITIZED}"
+                echo "SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE=${CLONE_NAME_SANITIZED}" >> $GITHUB_ENV
```
philerooski, line 56: Nit: This variable name is a little confusing. Without context, I would think that this is _the_ Synapse data warehouse database (i.e., the dev database). Could we make it more explicit that it's a clone by naming it something like `SYNAPSE_DATA_WAREHOUSE_DATABASE_CLONE`?

CONTRIBUTING.md
jaymedina: > Is this a case where someone else's changes are pulled in before yours and you need to rebase to get an up to date clone?

Exactly! We want developers to be mindful of other PRs so that their schema changes don't clash with other changes.

.github/workflows/test_with_clone.yaml
jaymedina: > Does this trigger if the PR is created with the label already?

Good point, I also had this concern but from what I've read, CI jobs can't get triggered on a closed PR. To be safe, I'll add this as a test after this gets merged.

> Do I need to remove the label and add it again?

Yes, you would need to remove it and add it again. I'll add this as a note in the contributing doc.

CONTRIBUTING.md
jaymedina: I would say creating the draft PR isn't optional, since you would want the `schemachange` test to pass before marking your PR as ready for review. What I'll add is a separate section for "mark your PR as ready for review after tests pass".

.github/workflows/test_with_clone.yaml
jaymedina: Good catch! On second thought, I think it would. I've updated the logic so that this job only runs on open PRs: https://github.com/Sage-Bionetworks/snowflake/pull/78/commits/03dfc5ac8ae02b867c78c68f555b9b2e67434b5e

CONTRIBUTING.md
jaymedina: https://github.com/Sage-Bionetworks/snowflake/pull/78/commits/ab9297a712bf260fa53f7e5dc247fc1acd85a2c8

.github/workflows/test_with_clone.yaml
jaymedina: https://github.com/Sage-Bionetworks/snowflake/pull/78/commits/ab9297a712bf260fa53f7e5dc247fc1acd85a2c8

.github/workflows/test_with_clone.yaml
jaymedina: The name convention needs to be this way because of how we run `schemachange`:

* If you look at the `schemachange` command, it points to the `synapse_data_warehouse` folder for where the [config file](https://github.com/Sage-Bionetworks/snowflake/blob/dev/synapse_data_warehouse/schemachange-config.yml) lives.
* This config file pulls the `database_name` from the environment variable `SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE`

In other words, if we change the variable to something else, `schemachange` will report back with an error saying it could not find the `SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE` environment variable.

.github/workflows/test_with_clone.yaml
```diff
@@ -0,0 +1,111 @@
...
+              shell: bash
+              run: |
+                  pip install schemachange==3.6.1
+                  pip install numpy==1.26.4
+                  pip install pandas==1.5.3
+
+            - name: Install SnowSQL
+              run: |
+                curl -O https://sfc-repo.snowflakecomputing.com/snowsql/bootstrap/1.2/linux_x86_64/snowsql-1.2.9-linux_x86_64.bash
+                SNOWSQL_DEST=~/bin SNOWSQL_LOGIN_SHELL=~/.profile bash snowsql-1.2.9-linux_x86_64.bash
```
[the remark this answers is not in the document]
philerooski: If we don't fix this in this PR, let's open a ticket to upgrade to using the action/latest major version of CLI.

CONTRIBUTING.md
philerooski: I think it would be more clear if we rephrase this:

> there may be other developers working on their own version of the development database

to something like this:

> there may have been changes made to the development database since your last clone.

And either rephrase or remove this

> Keep an eye out for other PRs with the `create_clone_and_run_schemachange` label

Since whether other people are also creating clones of the development database is a nonissue, it merely depends on whether there have been upstream changes to the development database.

.github/workflows/test_with_clone.yaml
philerooski: This has the potential to cause confusion if the clone of my PR does not reflect the HEAD of my feature branch. My preferred implementation would be that new commits to my branch also triggers the clone, if my branch meets the other conditions (PR/labeled).

CONTRIBUTING.md
philerooski: Ah, okay makes sense. In RECOVER we've been doing this a bit differently. Tests run as part of a regular PR (no need to remember to convert to a draft) and if the tests fail, then there's an implicit understanding that a review isn't needed yet. Although, tests failing as part of a PR is pretty rare -- since we run tests whenever there is a new commit to a feature branch -- issues tend to get caught before a PR is created / a review is needed.

.github/workflows/test_with_clone.yaml
philerooski: Hmm, yeah this seems to be a potential semantics issue with the way we've set up the schemachange config. To make the config more general that variable really ought to be named something like `TARGET_DATABASE`. But since the config was set up with the assumption that we would only ever be making changes to a specific database (dev/prod Synapse data warehouse) it has a very specific name.

But it is what it is. Not worth refactoring at this point imo.

CONTRIBUTING.md
jaymedina: > Tests run as part of a regular PR (no need to remember to convert to a draft)

Makes sense! Technically, the job can run on a non-draft PR, but it's good practice to mark PRs as drafts when they're not in their final state (all tests passing, etc), so I think we should enforce this in the workflow.

.github/workflows/test_with_clone.yaml
jaymedina: > This has the potential to cause confusion if the clone of my PR does not reflect the HEAD of my feature branch.

I would argue that this shouldn't cause confusion since you can match up the "last updated" on your DB to when you made your last commit, but I see this point.

If we go with that implementation, we may as well trigger the job for every commit regardless of whether it's labeled or not. We can revisit that discussion here. You mentioned that your CI pipeline for RECOVER triggers a zero copy clone for every commit, right? Did you experience any technical issues for edge cases where commits are made before the previous run was complete?

philerooski: If we had continued Snowflake development for RECOVER we probably would have used CLONE in some capacity. Instead, we had a stored procedure for each table that could be manually invoked to load data into that table, and a SQL script to do this in bulk for every data type, although this script wasn't used in our CI/CD. Snowflake stuff for RECOVER was never used in a production capacity, it was primarily a proof of concept. 

> I'm still on the fence with that idea

How come?

.github/workflows/test_with_clone.yaml
thomasyu888: * Re the loading of source tables: this is ok, the dev warehouse has small enough data that if we were ever to mess with the raw snapshot scheduled task pipeline, it would be ok to pull in more data
* Re the

.github/workflows/test_with_clone.yaml
thomasyu888: If we follow the schemachange workflow, the clone is mainly a space for you to test things manually.

Then once you know that works, ideally, transferring it here _could_ deploy the same code via schemachange to your cloned db, but may not be totally necessary in the short term. (There are some nuances of how to format schemachange scripts ofcourse)

Then merge into dev which then deploys it to dev.

.github/workflows/test_with_clone.yaml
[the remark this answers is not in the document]
thomasyu888: I vote to take the time to upgrade to the latest version while we are already on this task - thanks for raising that Phil!

.github/workflows/test_with_clone.yaml
jaymedina: > How come?

My main concern is the risk of running the job for a commit before the previous job has finished. In other words: back-to-back commits causing cloned DB conflicts, or clashing between R scripts that are running against the same clone.

I think you bring a valid point re: manually triggering the job leads to version inconsistency between the feature branch and the cloned DB, so I'm open to doing the run-for-every-commit implementation. But I need to do a test on this PR for the edge-cases I'm worried about, and then we can move forward from there. @philerooski

.github/workflows/test_with_clone.yaml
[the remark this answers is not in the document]
jaymedina: Updated!

jaymedina:
Hey all, just keeping you in the loop with this Snowflake devops stuff:

As I was working on testing an edge case for this PR, I realized there's an issue with using different roles for each step in the CI/CD pipeline. This is how it's working right now:
* I zero-copy clone the database via the DATA_ENGINEER role since that is the default role that DPE devs have access to (code) and they should be able to see their changes on Snowsight
* I deploy `schemachange` updates via the SYSADMIN role since DATA_ENGINEER doesn't have permissions to modify some of the existing tables (code)

So while I'm in DATA_ENGINEER, I can see my cloned DB, but I cannot see my dummy table that I made to test out the R scripts because it was made with SYSADMIN.

@thomasyu888, would it be possible for you to make a separate role for CI/CD work instead of the pipeline using DATA_ENGINEER and SYSADMIN for the two separate steps? This role would:
* Inherit the access privileges of SYSADMIN (so that schemachange can run without getting errors like [these](https://github.com/Sage-Bionetworks/snowflake/pull/78/commits/d0880d8f9042354f1c65da6ac549e8f483f09749))
* Be accessible to DPE engineers so they can view their `schemachange` deployment on their cloned DB on Snowsight

A related question:
I'm assuming the assets currently on Snowflake were made by running `schemachange` via the SYSADMIN role, correct? I'm able to see these assets while on DATA_ENGINEER, so why is this an issue suddenly?

cc @philerooski

philerooski:
@thomasyu888 @jaymedina 

Snowflake has a [suggested framework](https://docs.snowflake.com/en/user-guide/security-access-control-overview#role-hierarchy-and-privilege-inheritance) for organizing roles which [we will adopt](https://sagebionetworks.jira.com/browse/SNOW-113) at some point in the near future. As far as Snowflake objects are concerned, `SYSADMIN` ought to inherit privileges on _all_ objects. So a role which is responsible for running our CI/CD and inherits privileges from `SYSADMIN` wouldn't be possible within this framework. 

`DATAENGINEER` would be one of the Custom Roles in the framework. What I'm wondering: what makes `DATAENGINEER` less privileged than `SYSADMIN`? Which objects privileges in our account do we want to reserve only for SYSADMIN?

To the point -- whichever role we use for CI/CD, it ought to have privileges on every object in the dev Synapse data warehouse database. This might be `SYSADMIN`, or it could be a more precisely scoped role (`SYSADMIN_DEV`? I'm open to better ideas, or just using `SYSADMIN`). The cleanest way to then grant `DATAENGINEER` privileges to the objects which the CI/CD role has created is to create and grant a `DATABASE ROLE` specific to the cloned database to `DATAENGINEER`. (Do we automatically get a clone of the Synapse dev database's `DATABASE ROLE`, if it exists?).

Here's one possible implementation:
snowflake_cicd_roles

thomasyu888:
> Snowflake has a [suggested framework](https://docs.snowflake.com/en/user-guide/security-access-control-overview#role-hierarchy-and-privilege-inheritance) for organizing roles which [we will adopt](https://sagebionetworks.jira.com/browse/SNOW-113) at some point in the near future. As far as Snowflake objects are concerned, `SYSADMIN` ought to inherit privileges on _all_ objects. So a role which is responsible for running our CI/CD and inherits privileges from `SYSADMIN` wouldn't be possible within this framework.

@philerooski just a note here - I did that role refactor based off of your learnings a little bit ago - and it _is_ a lot cleaner.

image

jaymedina:
Yes, as a temporary solution, I can implement suggestions 3 and 2 so we can use `DPE_ENGINEER` for the schemachange execution. I'll be adding this in the next commit.

philerooski:
> I did that role refactor based off of your learnings a little bit ago

Should the ticket be closed or modified, then?

> during cloning, we could add a grant statement to GRANT ALL FUTURE DYNAMIC TABLES TO ROLE DATA ENGINEER. Thoughts?

What is it about dynamic tables specifically that requires this statement? Do we need FUTURE grants for other types of objects?

My main concern with the proposed workaround is that `DATA_ENGINEER` is quickly becoming a behemoth with no clear delineation between itself and `SYSADMIN`. But if we're in agreement that this is a temporary solution then I'm fine with proceeding.

jaymedina:
> My main concern with the proposed workaround is that DATA_ENGINEER is quickly becoming a behemoth with no clear delineation between itself and SYSADMIN

I agree with this. I don't think we should permanently grant the engineer role SYSADMIN privileges. A heads up: I'm fiddling around with a temporary solution for this until the role re-org. My commits will show me granting privileges to DPE_ENGINEER, but I'm being mindful that these privileges will _only_ exist within the scope of the cloned DB, and not within the actual dev and prod databases.

thomasyu888:
> My main concern with the proposed workaround is that DATA_ENGINEER is quickly becoming a behemoth with no clear delineation between itself and SYSADMIN

@philerooski / @jaymedina , I think there should be another ticket to fully determine whether the DATA_ENGINEER role should just be the OWNER of `synapse_data_warehouse` and `synapse_data_warehouse_dev`.  This way, the data_engineer role will be able to do all operations if needs to within those databases.

Similarly, since the `DATA_ENGINEER` role is rolled up into `SYSADMIN`, `SYSADMIN` should be able to do everything the `DATA_ENGINEER` role can do.  Note: I'm not proposing `DATA_ENGINEER` have the same level of permissions as `SYSADMIN`, but what I'm saying is the `DATA_ENGINEER` role _could_ have all the privileges the `SYSADMIN` role currently has WITHIN those two databases and clones

^ That said, I'm not tied to that, and I'll let Phil decide when we get to that design ticket.

.github/workflows/test_with_clone.yaml
```diff
@@ -0,0 +1,179 @@
...
+                # Transfer ownership of: schemas
+                snow sql -q "GRANT OWNERSHIP ON SCHEMA ${SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE}.SYNAPSE_RAW TO ROLE DATA_ENGINEER REVOKE CURRENT GRANTS;"
+                snow sql -q "GRANT OWNERSHIP ON SCHEMA ${SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE}.SYNAPSE TO ROLE DATA_ENGINEER REVOKE CURRENT GRANTS;"
+                
+                # Transfer ownership of: tables
+                snow sql -q "GRANT OWNERSHIP ON ALL TABLES IN SCHEMA ${SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE}.SYNAPSE_RAW TO ROLE DATA_ENGINEER REVOKE CURRENT GRANTS;"
+                snow sql -q "GRANT OWNERSHIP ON ALL TABLES IN SCHEMA ${SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE}.SYNAPSE TO ROLE DATA_ENGINEER REVOKE CURRENT GRANTS;"
+                
+                # Transfer ownership of: dynamic tables
+                snow sql -q "GRANT OWNERSHIP ON ALL DYNAMIC TABLES IN SCHEMA ${SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE}.SYNAPSE TO ROLE DATA_ENGINEER REVOKE CURRENT GRANTS;"
```
philerooski, line 109: Do we want to COPY CURRENT GRANTS rather than REVOKE so that we can imitate the source object as closely as possible?

jaymedina: Initially, I would say no, mainly because the only role that should be interacting with the clone and its objects is `DATA_ENGINEER` for the purpose of testing and validation anyway. Also because `COPY CURRENT GRANTS` would copy over the `DATA_ENGINEER`s previous grants before it became owner, so it would be redundant.

philerooski: I'm wondering how we might test changes to privileges. If we've already revoked current grants on an object, then how can we check that whatever changes we've made to privileges granted on an object reflect what we are expecting?

For example:

## `object_dev`

* (ownership privilege)
* privilege a
* privilege b
* privilege c

## `object_clone` (before schemachange)

* (ownership privilege)

## `object_clone` (after schemachange)

* (ownership privilege)
* privilege d

Assuming that schemachange applied some arbitrary change set to the privileges (like adding privilege d, revoking some other privileges), how can we test `grants on object_dev` + `our change set` = `grants we expect on this object`?

jaymedina: This is a good point - If we plan to add tests for updates in object privileges, then I agree we should not revoke any grants. I'll do a test to make sure `COPY CURRENT GRANTS` doesn't override `DATA_ENGINEER`s new ownership, and make a commit for this. Thanks!

jaymedina:
Something to note: 

The clone is built off the latest `dev` DB because this DB has schema changes that `prod` may not, **HOWEVER** the newest`snapshots` data flows directly into prod (see comment [here](https://github.com/Sage-Bionetworks/snowflake/pull/83#discussion_r1842624192)). My thinking is the data should ingest into `dev` before `prod` to make sure that the dynamic table queries (for example) in `dev` get updated and reflect the data as expected.

jaymedina:
From the conversation [here](https://sagebionetworks.slack.com/archives/C0AFCTGEQ/p1731611227968169), we can clone off of the `staging` DB (when it's created) instead of the `dev` DB

thomasyu888:
> From the conversation [here](https://sagebionetworks.slack.com/archives/C0AFCTGEQ/p1731611227968169), we can clone off of the `staging` DB (when it's created) instead of the `dev` DB

@jaymedina let's hold off on making changes until we can get all get our bearings on this.  Ideally the `dev` snapshots provide enough information for us to do our transforms, but in the incidences that it doesn't, you're right we do need to leverage staging.

The point of cloning DEV is so that the queries run faster due to it having less data (but similar structure) as the prod data

.github/workflows/test_with_clone.yaml
```diff
@@ -0,0 +1,179 @@
...
+      runs-on: ubuntu-latest
+      if: github.event.pull_request.merged == true || github.event.action == 'closed'
+      environment: dev
+      env:
+          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWSQL_PWD }}
+          SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWSQL_ACCOUNT }}
+          SNOWFLAKE_USER: ${{ secrets.SNOWSQL_USER }}
+          SNOWFLAKE_WAREHOUSE: ${{ secrets.SNOWSQL_WAREHOUSE }}
+          SNOWFLAKE_CLONE_ROLE: DATA_ENGINEER
+          CLONE_NAME: "${{ vars.SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE }}_${{ github.head_ref }}"
```
BryanFauble, line 132: If I am understanding this will only drop the clone when the PR is merged or closed. However, a cloned DB is created on any push when the PR is labeled and open.

Would this mean that only the last `github.head_ref` is cleaned up? In the end we would want to cleanup both the `head_ref` when things are merged, AND, when a new push is added (and head_ref is updated) we'd also want to drop the previous cloned DB.

jaymedina: > Would this mean that only the last github.head_ref is cleaned up?

Good question - Yes, this means that the  last `github.head_ref` is cleaned up. However, this is not a concern because it's the _only_ cloned DB on snowflake, for a given feature branch. All predecessor clones just get replaced by a new clone.

So for example if `head_ref` is `snow-90-my-branch`, the procedure goes like this:

1. an initial commit is made + job is triggered
2. new clone is made named: `synapse_data_warehouse_dev_snow_90_my_branch`
3. schemachange runs
4. another commit is made
5. **old clone is replaced by the new clone of `synapse_data_warehouse_dev` with a `CREATE OR REPLACE`. the new clone has the same name as the last clone, so it does a `REPLACE`**
6. schemachange runs
7. rinse and repeat
8. when it's time to merge, the clone is dropped (if it still exists). the drop clone job assumes the name hasn't been changed since its initial create, so this is a potential edge case actually, but it's probably enough to just say "don't change the clone name" in the CONTRIBUTIONS doc

philerooski:
@jaymedina I recommend closing this PR and opening another once you're ready for review. There's a lot of stale discussion on this PR.

PR #79, opened 2024-10-23
[SNOW-135] Add query to extract unique downloads per email domain

danlu1:
Problem:

Request to extract unique file downloads per email domain

Solution:

Draft query is added.


analytics/file_download_per_email_domain.sql
```diff
@@ -0,0 +1,30 @@
...
+
+
+WITH DEDUP_FILEHANDLE AS (
+    SELECT DISTINCT
+        USER_ID,
+        FILE_HANDLE_ID AS FD_FILE_HANDLE_ID,
+        RECORD_DATE,
+        PROJECT_ID
+    FROM
+        SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILEDOWNLOAD
```
thomasyu888, line 15: Nit: just a note here, you can explore the "association_object_type" column.  That will tell you "what" the downloads are associated it. You'll see something like this:
image

One thing we need to figure out at Sage is _what_ we want to consider as a "download"

analytics/file_download_per_email_domain.sql
```diff
@@ -0,0 +1,30 @@
...
+)
+
+SELECT
+    REGEXP_SUBSTR(USERS.EMAIL, '[^@]+$') as EMAIL_DOMAIN, 
+    count(*) AS NUMBER_OF_DOWNLOADS
```
thomasyu888, line 20: Nit: One thing you could add is `min(record_date)` and `max(record_date)` to always get the range.  That said it's not necessary

thomasyu888 (approved):
🔥 LGTM! Going to pre-approve!

PR #81, opened 2024-11-08
[SNOW-29] New dynamic table: `verificationsubmission_latest`

jaymedina:
## problem

The latest table for `verificationsubmissionsnapshots` does not exist in Snowflake.

Associated with #83

## solution

Introduce a new R script that updates a dynamic table showing the latest version of `verificationsubmissionsnapshots`

## testing

See #83


jaymedina:
Closing because we are moving away from using R scripts to alter dynamic tables.

PR #82, opened 2024-11-08
[SNOW-27] add version script to populate usergroup_latest table

danlu1:
Problem:

USERGROUP_LATEST table is not populated in the database.

Solution(WIP):

Add dynamic table USERGROUP_LATEST via a version script to pull deduplicated data from USERGROUPSNAPSHOT table. 
 
Test:

Tested in my personal stage data warehouse:
1. created `USERGROUP_LATEST` table and check for the ID `3576543` and get one row. 
Screenshot 2024-11-08 at 11 30 15 AM
2. Modified the base table `USERGROUPSNAPSHOT` by inserting two rows with newer `CHANGETIMESTAMP`
Screenshot 2024-11-08 at 12 20 00 PM
3. Refreshed the `USERGROUP_LATEST` table and got the updated row  for the ID `3576543`.
Screenshot 2024-11-08 at 1 30 42 PM

### DIFF
--- a/synapse_data_warehouse/synapse/dynamic_tables/R__populate_usergroup_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/R__populate_usergroup_latest.sql
@@ -1,0 +1,21 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE OR REPLACE DYNAMIC TABLE USERGROUP_LATEST
+    TARGET_LAG = DOWNSTREAM
+    WAREHOUSE = compute_xsmall
+    AS
+        WITH dedup_usergroup AS (
+            SELECT
+                *,
+                "row_number"()
+                    OVER (
+                        PARTITION BY ID
+                        ORDER BY CHANGE_TIMESTAMP DESC, SNAPSHOT_TIMESTAMP DESC
+                    )
+                    AS N
+            FROM SYNAPSE_DATA_WAREHOUSE_DanLu_STAGE.SYNAPSE_RAW.USERGROUPSNAPSHOTS --noqa: TMP
+            QUALIFY
+                N=1
+        )
+        SELECT * EXCLUDE N
+        FROM dedup_usergroup;
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__usergroup_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__usergroup_latest.sql
@@ -15,7 +15,7 @@ CREATE OR REPLACE DYNAMIC TABLE USERGROUP_LATEST
                     AS N
             FROM {{database_name}}.SYNAPSE_RAW.USERGROUPSNAPSHOTS --noqa: TMP
             WHERE 
-                (SNAPSHOT_DATE >= CURRENT_TIMESTAMP - INTERVAL '14 days') 
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '14 days'
             QUALIFY
                 N=1
         )
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__usergroup_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__usergroup_latest.sql
@@ -1,7 +1,7 @@
 use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
 
 CREATE OR REPLACE DYNAMIC TABLE USERGROUP_LATEST
-    TARGET_LAG = DOWNSTREAM
+    TARGET_LAG = '1 day'
     WAREHOUSE = compute_xsmall
     AS
         WITH dedup_usergroup AS (
@@ -13,7 +13,9 @@ CREATE OR REPLACE DYNAMIC TABLE USERGROUP_LATEST
                         ORDER BY CHANGE_TIMESTAMP DESC, SNAPSHOT_TIMESTAMP DESC
                     )
                     AS N
-            FROM SYNAPSE_DATA_WAREHOUSE_DanLu_STAGE.SYNAPSE_RAW.USERGROUPSNAPSHOTS --noqa: TMP
+            FROM {{database_name}}.SYNAPSE_RAW.USERGROUPSNAPSHOTS --noqa: TMP
+            WHERE 
+                (SNAPSHOT_DATE >= CURRENT_TIMESTAMP - INTERVAL '14 days') 
             QUALIFY
                 N=1
         )


danlu1:
@thomasyu888 Right now, I set `target_lag` to DOWNSTREAM for testing purpose. I saw team_latest used `1 day`. What is the maximum amount of time that we accept the Latest table’s content to lag behind to the snapshot table?

synapse_data_warehouse/synapse/dynamic_tables/R__populate_usergroup_latest.sql
```diff
@@ -0,0 +1,21 @@
...
+        WITH dedup_usergroup AS (
+            SELECT
+                *,
+                "row_number"()
+                    OVER (
+                        PARTITION BY ID
+                        ORDER BY CHANGE_TIMESTAMP DESC, SNAPSHOT_TIMESTAMP DESC
+                    )
+                    AS N
+            FROM SYNAPSE_DATA_WAREHOUSE_DanLu_STAGE.SYNAPSE_RAW.USERGROUPSNAPSHOTS --noqa: TMP
```
thomasyu888, line 16: I would set a snapshot_date timeline to reduce the number of "user groups" that should be deleted.  (2 weeks) is a good timeframe

synapse_data_warehouse/synapse/dynamic_tables/R__populate_usergroup_latest.sql
```diff
@@ -0,0 +1,21 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE OR REPLACE DYNAMIC TABLE USERGROUP_LATEST
```
thomasyu888, line 3: Can we see if there is `CREATE OR ALTER` command for dynamic tables?

thomasyu888:
> @thomasyu888 Right now, I set `target_lag` to DOWNSTREAM for testing purpose. I saw team_latest used `1 day`. What is the maximum amount of time that we accept the Latest table’s content to lag behind to the snapshot table?

I would take a look at these docs: https://docs.snowflake.com/en/user-guide/dynamic-tables-refresh.  and get an understanding of downstream does.  Let me know if you think we should use downstream or target lag based on the scenarios - happy to discuss.

synapse_data_warehouse/synapse/dynamic_tables/R__populate_usergroup_latest.sql
danlu1: As the [doc](https://docs.snowflake.com/en/sql-reference/sql/create-table#label-create-or-alter-table-usage-notes:~:text=Currently%20only%20supports%20permanent%2C%20temporary%2C%20and%20transient%20tables.%20Read%2Donly%2C%20external%2C%20dynamic%2C%20Apache%20Iceberg%E2%84%A2%2C%20and%20hybrid%20tables%20are%20not%20supported.): `CREATE OR ALTER` currently only supports permanent, temporary, and transient tables. Read-only, external, dynamic, Apache Iceberg™, and hybrid tables are not supported.

synapse_data_warehouse/synapse/dynamic_tables/R__populate_usergroup_latest.sql
thomasyu888, line 16:
```suggestion
            FROM {{database_name}}.SYNAPSE_RAW.USERGROUPSNAPSHOTS --noqa: TMP
```

danlu1: Thanks for catching this.

danlu1:
> I would take a look at these docs: https://docs.snowflake.com/en/user-guide/dynamic-tables-refresh. and get an understanding of what DOWNSTREAM does. Let me know if you think we should use downstream or target lag based on the scenarios - happy to discuss.

The DOWNSTREAM target_lag means that "the dynamic table should refresh on demand when other dependent dynamic tables refresh. This refresh can be triggered by a manual or scheduled refresh of a downstream dynamic table.". In this case, since there is no downstream tables, I can manually refresh this dynamic table using 
```
ALTER DYNAMIC TABLE SYNAPSE_DATA_WAREHOUSE_DanLu_STAGE.synapse.USERGROUP_LATEST REFRESH;
```
In practice, once the dynamic table is created, it should update itself within a preset lag of time once the base table is changed.

thomasyu888:
> In practice, once the dynamic table is created, it should update itself within a preset lag of time once the base table is changed.

Hmm... we probably don't want to have to do this manually.

```
> ALTER DYNAMIC TABLE SYNAPSE_DATA_WAREHOUSE_DanLu_STAGE.synapse.USERGROUP_LATEST REFRESH;
```

I think from the documentation, the DT is also not updated if the base table has been updated unless we do `TARGET_LAG` so we probably _want_ a TARGET_LAG to maybe an hour instead of a 1 day so that it's refreshed.  What I wanted to avoid was this command running every hour - which isn't clear to me in the documentation.

synapse_data_warehouse/synapse/dynamic_tables/R__populate_usergroup_latest.sql
thomasyu888, line 1: @danlu1 I'd update this to reflect what we discussed in the other PR

danlu1: Done.

thomasyu888 (approved):
🔥 LGTM!  Thanks for doing this!

synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__usergroup_latest.sql
```diff
@@ -0,0 +1,23 @@
...
+                *,
+                "row_number"()
+                    OVER (
+                        PARTITION BY ID
+                        ORDER BY CHANGE_TIMESTAMP DESC, SNAPSHOT_TIMESTAMP DESC
+                    )
+                    AS N
+            FROM {{database_name}}.SYNAPSE_RAW.USERGROUPSNAPSHOTS --noqa: TMP
+            WHERE 
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '14 days'
```
thomasyu888, line 18: Whew, good catch !

synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__usergroup_latest.sql
```diff
@@ -0,0 +1,23 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE OR REPLACE DYNAMIC TABLE USERGROUP_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+        WITH dedup_usergroup AS (
+            SELECT
+                *,
+                "row_number"()
```
philerooski, line 10: Why did you put "row_number" in quotes?

danlu1: I borrowed this from this R__populate_team_latest.sql. I think this is not necessary though since ROW_NUMBER is default function. But @jaymedina do you know why quotation is used in this case?

jaymedina: No particular reason, but yes I think we should stick to `ROW_NUMBER` to be consistent. This is from an older query that I migrated out into its own script.

danlu1: @jaymedina Thanks so much for the clarification. @philerooski do you want me to update it in another PR?

thomasyu888: Let's not worry about it right now since it's already been committed, next time we use this function, we can be consistent. Thanks everyone for catching this!

philerooski: Thanks, I was just curious. It only affects readability.

danlu1: Sounds great. Thanks everyone!

PR #83, opened 2024-11-12
[SNOW-29] Initialize dynamic table: `verificationsubmission_latest`

jaymedina:
## problem

* The latest table for `verificationsubmissionsnapshots` does not exist in Snowflake.
* Following versioning best practices for Schemachange, we should initialize the dynamic table from a separate V script to mark its introduction into the system.

Associated with #81.

## solution

Add a V script to initialize `verificationsubmission_latest`

## testing

There are 19 unique submissions stored in the snapshots table, meaning there should be 19 total rows in the deduplicated latest table. The reason being that each submission can only be in 1 state - CREATED or UPDATED:

image

We verify that the V script generates a table with 19 rows:

image

We verify that the 19 unique IDs in the snapshots table matches the 19 unique IDs in the latest table:

image

( This query would produce results if there were any missing IDs from either table )

We use a few IDs to manually verify that the latest state of each submission is reflected in the latest table:

- [x] 67

image
image

- [x] 65

image
image

- [x] 36

image
image

### DIFF
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
@@ -4,25 +4,11 @@ CREATE DYNAMIC TABLE IF NOT EXISTS VERIFICATIONSUBMISSION_LATEST
     TARGET_LAG = '1 day'
     WAREHOUSE = compute_xsmall
     AS
-    WITH latest_rows AS (
-        SELECT
-            id AS latest_id,
-            MAX(snapshot_timestamp) AS latest_timestamp
-        FROM
-            verificationsubmissionsnapshots
-        GROUP BY
-            latest_id
-    ),
-    latest_unique_rows AS (
+    WITH latest_unique_rows AS (
         SELECT
             verificationsubmissionsnapshots.*,
         FROM
-            verificationsubmissionsnapshots
-        JOIN
-            latest_rows
-        ON
-            verificationsubmissionsnapshots.id = latest_rows.latest_id
-            AND verificationsubmissionsnapshots.snapshot_timestamp = latest_rows.latest_timestamp
+            {{database_name}}.synapse_raw.verificationsubmissionsnapshots --noqa: TMP
         QUALIFY ROW_NUMBER() OVER (
                 PARTITION BY id
                 ORDER BY snapshot_timestamp DESC
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
@@ -1,6 +1,6 @@
 use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
 
-CREATE OR REPLACE DYNAMIC TABLE VERIFICATIONSUBMISSION_LATEST
+CREATE DYNAMIC TABLE IF NOT EXISTS VERIFICATIONSUBMISSION_LATEST
     TARGET_LAG = '1 day'
     WAREHOUSE = compute_xsmall
     AS
@@ -17,7 +17,7 @@ CREATE OR REPLACE DYNAMIC TABLE VERIFICATIONSUBMISSION_LATEST
         SELECT
             verificationsubmissionsnapshots.*,
             ROW_NUMBER() OVER (
-                PARTITION BY snapshot_timestamp
+                PARTITION BY id
                 ORDER BY snapshot_timestamp DESC
             ) AS row_num
         FROM
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
@@ -6,12 +6,12 @@ CREATE OR REPLACE DYNAMIC TABLE VERIFICATIONSUBMISSION_LATEST
     AS
     WITH latest_rows AS (
         SELECT
-            verificationsubmissionsnapshots.id AS the_id,
+            id AS latest_id,
             MAX(snapshot_timestamp) AS latest_timestamp
         FROM
             verificationsubmissionsnapshots
         GROUP BY
-            verificationsubmissionsnapshots.id
+            latest_id
     ),
     latest_unique_rows AS (
         SELECT
@@ -25,11 +25,11 @@ CREATE OR REPLACE DYNAMIC TABLE VERIFICATIONSUBMISSION_LATEST
         JOIN
             latest_rows
         ON
-            verificationsubmissionsnapshots.id = latest_rows.the_id
+            verificationsubmissionsnapshots.id = latest_rows.latest_id
             AND verificationsubmissionsnapshots.snapshot_timestamp = latest_rows.latest_timestamp
     )
     SELECT
-        *
+        * EXCLUDE (row_num)
     FROM
         latest_unique_rows
     WHERE
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
@@ -11,6 +11,8 @@ CREATE DYNAMIC TABLE IF NOT EXISTS VERIFICATIONSUBMISSION_LATEST
             verificationsubmissionsnapshots.*,
         FROM
             {{database_name}}.synapse_raw.verificationsubmissionsnapshots --noqa: TMP
+        WHERE
+            SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '14 DAYS'
         QUALIFY ROW_NUMBER() OVER (
                 PARTITION BY id
                 ORDER BY change_timestamp DESC, snapshot_timestamp DESC
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
@@ -33,4 +33,4 @@ CREATE DYNAMIC TABLE IF NOT EXISTS VERIFICATIONSUBMISSION_LATEST
     FROM
         latest_unique_rows
     ORDER BY
-        latest_unique_rows.id;
+        latest_unique_rows.id ASC;
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
@@ -16,10 +16,6 @@ CREATE DYNAMIC TABLE IF NOT EXISTS VERIFICATIONSUBMISSION_LATEST
     latest_unique_rows AS (
         SELECT
             verificationsubmissionsnapshots.*,
-            ROW_NUMBER() OVER (
-                PARTITION BY id
-                ORDER BY snapshot_timestamp DESC
-            ) AS row_num
         FROM
             verificationsubmissionsnapshots
         JOIN
@@ -27,12 +23,14 @@ CREATE DYNAMIC TABLE IF NOT EXISTS VERIFICATIONSUBMISSION_LATEST
         ON
             verificationsubmissionsnapshots.id = latest_rows.latest_id
             AND verificationsubmissionsnapshots.snapshot_timestamp = latest_rows.latest_timestamp
+        QUALIFY ROW_NUMBER() OVER (
+                PARTITION BY id
+                ORDER BY snapshot_timestamp DESC
+            ) = 1
     )
     SELECT
-        * EXCLUDE (row_num)
+        *
     FROM
         latest_unique_rows
-    WHERE
-        row_num = 1
     ORDER BY
         latest_unique_rows.id;
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
@@ -1,0 +1,38 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE OR REPLACE DYNAMIC TABLE VERIFICATIONSUBMISSION_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_rows AS (
+        SELECT
+            verificationsubmissionsnapshots.id AS the_id,
+            MAX(snapshot_timestamp) AS latest_timestamp
+        FROM
+            verificationsubmissionsnapshots
+        GROUP BY
+            verificationsubmissionsnapshots.id
+    ),
+    latest_unique_rows AS (
+        SELECT
+            verificationsubmissionsnapshots.*,
+            ROW_NUMBER() OVER (
+                PARTITION BY snapshot_timestamp
+                ORDER BY snapshot_timestamp DESC
+            ) AS row_num
+        FROM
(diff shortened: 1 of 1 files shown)


synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
```diff
@@ -0,0 +1,38 @@
...
+    WITH latest_rows AS (
+        SELECT
+            id AS latest_id,
+            MAX(snapshot_timestamp) AS latest_timestamp
+        FROM
+            verificationsubmissionsnapshots
+        GROUP BY
+            latest_id
+    ),
+    latest_unique_rows AS (
```
jaymedina, line 16: I chose to work with `ROW_NUMBER` instead of `DISTINCT` because the `DISTINCT` solution was leading to a duplicate row where there was a discrepancy in `CREATED_ON`. The timestamp for this column was off my milliseconds and created this duplicate:

image
 
My solution strictly enforces no-duplicates, but at the cost of dropping any duplicate silently. I'm opening this up for discussion in case we want to take a more cautious approach.

thomasyu888: Could you speak to why you didn't choose to partition by ID and order by snapshot date to get the ranking?

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
```diff
@@ -0,0 +1,38 @@
...
+        FROM
+            verificationsubmissionsnapshots
+        JOIN
+            latest_rows
+        ON
+            verificationsubmissionsnapshots.id = latest_rows.latest_id
+            AND verificationsubmissionsnapshots.snapshot_timestamp = latest_rows.latest_timestamp
+    )
+    SELECT
+        * EXCLUDE (row_num)
```
thomasyu888, line 32: Instead of this, look into the QUALIFY command in snowflake.

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
```diff
@@ -0,0 +1,38 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE OR REPLACE DYNAMIC TABLE VERIFICATIONSUBMISSION_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_rows AS (
+        SELECT
+            id AS latest_id,
+            MAX(snapshot_timestamp) AS latest_timestamp
```
thomasyu888, line 10: Is there a change timestamp in this table?

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
```diff
@@ -0,0 +1,38 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE OR REPLACE DYNAMIC TABLE VERIFICATIONSUBMISSION_LATEST
```
thomasyu888, line 3: No need to use V script, just use an R Script to do this.

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
thomasyu888: Hmm, this is where data size probably comes into the picture, have you taken a look at the prod tables?

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
jaymedina: This was a typo and does not reflect the final solution I have written on my SQL worksheet. Something weird must have happened in the copy/paste process. Thanks for pointing it out!

thomasyu888: A good indication of why your clone solution is a good future solution to work towards!

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
jaymedina: > Is there a change timestamp in this table?

If you're asking if `CHANGE_TIMESTAMP` is in the final table result from this query, yes it is!

jaymedina: Update: I've added `CHANGE_TIMESTAMP` in my CTE in [this commit](https://github.com/Sage-Bionetworks/snowflake/pull/83/commits/0fba27a378afdd1caad138fc2c305897fb23a26b).

See my comment [here](https://github.com/Sage-Bionetworks/snowflake/pull/84#discussion_r1840914306)

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
thomasyu888, line 1: Can we update this to an R script?

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
```diff
@@ -0,0 +1,24 @@
...
+    -- verification submission ID...
+    WITH latest_unique_rows AS (
+        SELECT
+            verificationsubmissionsnapshots.*,
+        FROM
+            {{database_name}}.synapse_raw.verificationsubmissionsnapshots --noqa: TMP
+        QUALIFY ROW_NUMBER() OVER (
+                PARTITION BY id
+                ORDER BY change_timestamp DESC, snapshot_timestamp DESC
+            ) = 1
```
thomasyu888, line 17: Can we add `snapshot_timestamp` within last 2 weeks?

thomasyu888 (commented):
Thanks for doing this, I'll leave it to @philerooski for a final review so he can get familiar with this

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
jaymedina: There's a [separate PR](https://github.com/Sage-Bionetworks/snowflake/pull/81) for the R script - I created a separate V script for this dynamic table to mark its introduction in the `change_history` table

thomasyu888: Hmm, here are my thoughts around this, if R scripts are just run regardless, then the github commit history and PRs can track the introduction of these scripts.

Thoughts?

jaymedina: yep sounds good. I don't see any huge benefit to doing it the other way. Leaving this open though in case @philerooski wants to chime in.

thomasyu888: I think we need to separate out concerns

1. The data that the dynamic table creates is always a total refresh from the raw table as it's written.  It's never an append.
2. The R Script allows us to keep a static state of the dynamic table but youre right it blows away the dynamic table object and the history of the operations in the old dynamic table.

Regarding #2, if we wanted to alter a dynamic table then - it would be a V script to say. ALTER DYNAMIC TABLE, since create or alter isn't supported. However since #1 is the case here, is it valuable to track all the times that the dynamic table ran?

jaymedina: > It's not an append (I think).

Correct, we do a Full refresh, meaning the entire dynamic table gets replaced each time. You can go on Snowsight and hover over this icon for the full explanation for each table.

image

> A dynamic table, unlike a view, does contain data that would be lost if it were replaced.

Since the dynamic table reflects the latest state of all the records in its `snapshots` counterpart, the data is never lost (it's just in `snapshots`).

> if we wanted to alter a dynamic table then - it would be a V script to say. ALTER DYNAMIC TABLE, since create or alter isn't supported. 

I would say that altering (appending to) existing dynamic tables instead of rebuilding them defeats the purpose of why we use those tables here. We want to delete any older states of a given record and replace it by its newest state, not keep the two states.

If we wanted to alter by adding/removing columns, we would just change the R script.

philerooski: > The data that the dynamic table creates is always a total refresh from the raw table as it's written. It's not an append (I think).

This is dependent on the [REFRESH_MODE](https://docs.snowflake.com/en/user-guide/dynamic-tables-refresh) of the dynamic table. Since the query used to construct this dynamic table doesn't use any of the [unsupported operators](https://docs.snowflake.com/en/user-guide/dynamic-tables-limitations#support-for-incremental-refresh) and does use [supported operators](https://docs.snowflake.com/en/user-guide/dynamic-tables-refresh#supported-queries-in-incremental-refresh), it seems very likely that this dynamic table would do an incremental refresh (the default `REFRESH_MODE` is `AUTO`).

If computing the dynamic table from scratch isn't very expensive, then I think it's fine to do a full refresh at the rate we merge PRs. We don't _need_ to optimize this aspect. But if we anticipate adding additional downstream tables which use this dynamic table as a base table (and which could benefit from the query caching/optimization that Snowflake does behind the scenes) then it may be better to keep things lean and work with the incremental refresh (i.e., make updates using `ALTER`).

IMO, for Snowflake objects which contain data we should try to maintain that data during updates if it's not too onerous.

danlu1: Thanks guys for this discussion. I was running into the same issue version_script vs. repeatable_script and  full refresh vs. incremental refresh mode in my [PR](https://github.com/Sage-Bionetworks/snowflake/pull/82). Just wanted to share some of my thoughts here for further discussion. 

For script type: 

I originally chose to use R script for usergroup_latest table since I think the logic may need regular updates. For example, we might want to tweak the time ranges. I didn't create a version scripts for the DT since it initializes on creation by default. 
However, after some researches(such as this [one](https://github.com/Snowflake-Labs/schemachange/issues/262)), I found that version script would be more appropriate for this case. Given the dynamic table itself will update in response to changes in the base table, we don't need a R script to be re-applied to make data consistent even though it's only executed if changes made to the scripts. Also, if changes to the table structure or CTE logic are needed, those can be handled by adding new version scripts in the future. 
If we really would like to implement R script, we need to make sure that `IF NOT EXISTS` is used as suggested in this [post](https://stackoverflow.com/questions/70114439/versioned-or-repeatable-scripts-for-sequence-partitions-flyway). 

For refresh mode, after adding the time range (where snapshot_date >= specific time period) to the query, the refresh mode automatically set by snowflake changed from incremental to full refresh. As Phil said, we might need to double check if we want to implement incremental refresh mode if there are downstream tables which use this dynamic table as a base table. As the snowflake doc, incremental refresh mode is preferred if changes in the source or target is less than 5% of the rows. Since refresh_mode is immutable after the dynamic table is created, I plan to evaluation the change rate in my base table to set the refresh mode in my script.

jaymedina: > What happens if you explicitly set the refresh mode to INCREMENTAL. Does the dynamic table fail to populate?

I'm not sure, but I can experiment with this off a branch that tracks my CI/CD pipeline branch. I'll tag you in it when I'm done.

> Also, if changes to the table structure or CTE logic are needed, those can be handled by adding new version scripts in the future.

Thanks Dan! I'm warming up to this and am in agreement. Handling the structural changes with V scripts makes any new changes feel more intentional, especially if the dynamic tables have downstream dependencies.

Regarding incremental vs. full refresh, I think this is something to investigate if we ever plan our tables to have a really short `TARGET_LAG` time (what I'm implying is that, for short lags, the time for a refresh runs the risk of exceeding the `TARGET_LAG`, and the actual lag ends up being longer than what's designed), but our current dynamic tables are designed to reflect ~2 weeks worth of rows, as such right now I'm seeing that the full refresh times for our dynamic tables are < 1 minute, so is it worth it to worry about this while we are designing our queries (at least for the 2-week latest tables)? I'm open to your thoughts.

image

To view the refresh times, go to one of the dynamic tables in our DEV database on Snowsight and click the **Refresh History** tab for one of the tables.

jaymedina: @philerooski if your dynamic table query is not supported by incremental refresh, but you enforce it, you end up with a [SQL compilation error](https://github.com/Sage-Bionetworks/snowflake/actions/runs/11841222994/job/32996943069?pr=86#step:11:73)

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
jaymedina: The latest snapshot taken for `verificationsubmissionsnapshots` was Jan 11, 2024 so this would create an empty table. Is there a planned schedule for more frequent updates to this snapshots table? And maybe we should also revisit how we determine the "latest" window for these tables (i.e. why is 2 weeks from today the standard that we usually set for these). I can make a ticket to discuss this later since we're already discussing the R vs V scripts here.

cc @philerooski

thomasyu888: The biggest issue is it's hard to test some of this stuff on the dev database because data doesn't flow in daily.

jaymedina: Hmm okay, that's good to know. This is a separate convo more related to the CI pipeline work, so I left a comment [there](https://github.com/Sage-Bionetworks/snowflake/pull/78#issuecomment-2477015135) for us to address when we get back to that

thomasyu888: If we do leave out the 2 week window, none of these latest tables are usable (imo) until we get a better understanding of the data within these tables (which require discussion with Platform)

jaymedina: > Do you have access to the prod snapshots?

Nope, you're right, I was looking at `dev` 😅 The latest snapshot timestamp for this table on `prod` is from Nov 14, 2024, so yes I can make this query for a 2-week window

synapse_data_warehouse/synapse/dynamic_tables/V2.21.0__verificationsubmission_latest.sql
philerooski: > I'm seeing that the full refresh times for our dynamic tables are < 1 minute, so is it worth it to worry about this while we are designing our queries (at least for the 2-week latest tables)?

No, it's not worth holding up the current `*_latest` PRs just to optimize the refresh method at this point. Probably, a full refresh will never be overly burdensome for these `*_latest` tables, although it's plausible that future tables which derive from these `*_latest` tables will perform expensive operations and could benefit from static result sets in the `*_latest` tables (i.e., `*_latest` uses an incremental refresh).

In any case, it doesn't even seem possible to do an incremental refresh using the current query as per Dan's example.

jaymedina: Yep this makes sense to me. What we can do is table the refresh optimization for "phase 2" i.e. after we finish introducing all of the latest tables, and before we start thinking about creating downstream dependencies to these tables.

For now, are we in agreement with moving forward with:
1. Introducing dynamic tables with V scripts
2. Altering existing dynamic tables with V scripts
3. Prioritizing query optimization (i.e. using non-deterministic functions etc) over refresh optimization (for now)
?

If so, I'll mark this thread as resolved and request a final review!

philerooski (approved):

PR #84, opened 2024-11-13
[SNOW-151] Initialize dynamic table: `accessrequirement_latest`

jaymedina:
## testing

### DIFF
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__accessrequirement_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__accessrequirement_latest.sql
@@ -1,0 +1,22 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE DYNAMIC TABLE IF NOT EXISTS ACCESSREQUIREMENT_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_unique_rows AS (
+        SELECT
+            accessrequirementsnapshots.*,
+        FROM
+            {{database_name}}.synapse_raw.accessrequirementsnapshots --noqa: TMP
+        QUALIFY ROW_NUMBER() OVER (
+                PARTITION BY id
+                ORDER BY change_timestamp DESC,snapshot_timestamp DESC
+            ) = 1
+    )
+    SELECT
+        *
+    FROM
+        latest_unique_rows
+    ORDER BY
+        latest_unique_rows.id ASC;


synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__accessrequirement_latest.sql
```diff
@@ -0,0 +1,22 @@
...
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_unique_rows AS (
+        SELECT
+            accessrequirementsnapshots.*,
+        FROM
+            {{database_name}}.synapse_raw.accessrequirementsnapshots --noqa: TMP
+        QUALIFY ROW_NUMBER() OVER (
+                PARTITION BY id
+                ORDER BY change_timestamp DESC,snapshot_timestamp DESC
```
jaymedina, line 14: By relying solely on `snapshot_timestamp` to get the latest state of an item (in this case AR configuration), we run the risk of not getting the latest state **because `snapshot_timestamp` can also reflect the scheduled snapshots, not just the ones that are triggered from a change**.

Because of this, I'm also ordering the rows in the partitions by `change_timestamp` in addition to `snapshot_timestamp` as a precaution.

image

synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__accessrequirement_latest.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE DYNAMIC TABLE IF NOT EXISTS ACCESSREQUIREMENT_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_unique_rows AS (
+        SELECT
+            accessrequirementsnapshots.*,
+        FROM
+            {{database_name}}.synapse_raw.accessrequirementsnapshots --noqa: TMP
+        QUALIFY ROW_NUMBER() OVER (
```
thomasyu888, line 12: This is nice! Please do add in the snapshot_timestamp < 2 weeks check

jaymedina: I did not add the timestamp restriction because `accessrequirementsnapshots` records deletions:

image

For my understanding, we add these timestamp windows because:
* It reduces the scope of the query, making the CTE more efficient
* For tables that do not capture the deletion of a record, it reduces the chance that a record shows up in `latest` even though it's deleted from the system. For example, if a record was deleted but `change_type = 'DELETE'` is not supported in the table, setting the 2-week window minimizes the chance of it showing up in the final table, assuming the record wasn't deleted within those 2 weeks.

Let me know if this is correct

thomasyu888: Ooo thanks for checking! This then becomes a bit more complicated! 

How do we use those DELETE events to make sure we are removing the ones that should be removed?

thomasyu888: Ah I think 7 days may be too short because there might be records that cut in and out of the snapshot dates, that's why I chose 14 or in some cases 30.

I think that should work if you make sure N=1 isn't a DELETE event but we should test that theory.

I do like your thought of tracking delete events but thats not what's in the latest tables currently.

jaymedina: The delete events could also be in their own tables - I think that would make more sense.

So just to clarify once more:
* For snapshots that record DELETE events, we _don't_ need a timestamp window because we can filter out deleted items with `change_type`
* For snapshots that do not record DELETE events, we use a timestamp window to filter out deleted items, so long as the deletion happened outside of the timestamp window
* The other advantage is using a timestamp window improves query efficiency since you're pulling in less data in your CTE

So for this case, since I can filter out the deleted records using `change_type`, I don't _need_ to use a timestamp window, but I can use one for query efficiency.

thomasyu888: Great summary! Thanks!

jaymedina: perfect, thanks!

cc @philerooski @danlu1 so they know too

danlu1 (approved):
LGTM! Excellent work.

synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__accessrequirement_latest.sql
jaymedina: Updated https://github.com/Sage-Bionetworks/snowflake/pull/84/commits/3cbc2d6e1c9f2a921644b60369664b822565bbef

synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__accessrequirement_latest.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE DYNAMIC TABLE IF NOT EXISTS ACCESSREQUIREMENT_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_unique_rows AS (
+        SELECT
+            accessrequirementsnapshots.*,
+        FROM
+            {{database_name}}.synapse_raw.accessrequirementsnapshots --noqa: TMP
+        WHERE
```
thomasyu888, line 12: @jaymedina , I'll defer to @philerooski first to see if there are any issues with this query.  Hint: it's all about _when_ the DELETE event is removed.

Note, there may not be any issues, i just haven't dove into it.

synapse_data_warehouse/synapse/dynamic_tables/V2.22.0__accessrequirement_latest.sql
thomasyu888, line 1: Before merging this, we would need to change this to V2.23

jaymedina: whoops, thanks for the catch

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
```diff
@@ -0,0 +1,26 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE DYNAMIC TABLE IF NOT EXISTS ACCESSREQUIREMENT_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_unique_rows AS (
+        SELECT
+            accessrequirementsnapshots.*,
```
philerooski, line 9: Why is the table name included in the SELECT statement?

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
```diff
@@ -0,0 +1,26 @@
...
+        FROM
+            {{database_name}}.synapse_raw.accessrequirementsnapshots --noqa: TMP
+        WHERE
+            CHANGE_TYPE != 'DELETE'
+        AND
+            SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '14 DAYS'
```
philerooski, line 15: Snapshots are taken at least every 2 weeks?

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
```diff
@@ -0,0 +1,26 @@
...
+        QUALIFY ROW_NUMBER() OVER (
+                PARTITION BY id
+                ORDER BY change_timestamp DESC, snapshot_timestamp DESC
```
philerooski, line 18: Why do we order primarily by `change_timestamp`? Isn't `snapshot_timestamp` sufficient? Or could there be a case where a create/update/delete event occurred but there was no snapshot?

philerooski (approved):

thomasyu888 (changes requested):
@philerooski take a look at my most recent comment

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
```diff
@@ -0,0 +1,26 @@
...
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_unique_rows AS (
+        SELECT
+            accessrequirementsnapshots.*,
+        FROM
+            {{database_name}}.synapse_raw.accessrequirementsnapshots --noqa: TMP
+        WHERE
+            CHANGE_TYPE != 'DELETE'
```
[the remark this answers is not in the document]
philerooski: @thomasyu888 the evaluation ordering is in the docs: https://docs.snowflake.com/en/sql-reference/constructs/qualify

WHERE is evaluated before QUALIFY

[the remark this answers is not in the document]
thomasyu888: Ah so we have to change this SQL function to remove the change types after the qualify

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
jaymedina: That's a good point - there's no need to explicitly declare the table name since there are no other tables we're working with in this context. [Updated](https://github.com/Sage-Bionetworks/snowflake/pull/84/commits/4cd118e72c57d8d7c3a9a7316b632d659c05a41e)!

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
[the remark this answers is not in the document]
jaymedina: I see. Removing the delete events before getting the latest snapshots while still letting those assets get ranked by time is essentially treating them as if they're still there.

Changing the query to have the WHERE clause outside of the CTE led to 1 less row, the one below:
image

This row indeed have a delete event within the last 8 days (our new window), which wasn't captured in the original query
image

Good to note the order of these operations in the future. Thanks for the catch!

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
jaymedina: Edit: With the update [here](https://github.com/Sage-Bionetworks/snowflake/pull/84/commits/b204445e1f264c769afd6974b169231b42724522) I do see the same 3,803 assets using 7 days as using 14 days, but I'm just going to keep it at 8 days because the time window doesn't really matter in this query.

philerooski: Excellent. Do you know if there's documentation (from platform) about what is snapshotted and how frequently? Or is everything snapshotted every 7 days?

thomasyu888: Btw, platform themselves use 14 days in the de-duplication.  That is probably safer because we don't want to risk snapshotting not working.

We should ask them what the optimal interval is when de-dupping.

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
philerooski: Thanks, that makes sense.

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
jaymedina: > what is snapshotted and how frequently?

From a conversation with Marco and from [this documentation](https://sagebionetworks.jira.com/wiki/spaces/DW/pages/3034775570/Raw+tables+in+Athena+Snowflake#Important-terminology), it looks like snapshots happen:
1. every 7 days
2. every time an interaction with an asset happens, such that its "state" changes (e.g. an account verification submission is approved, an ACL configuration is deleted, etc)

> Btw, platform themselves use 14 days in the de-duplication.

Platform team does de-duplication before uploading the data to S3?

I'm fine with keeping it at 14 days if that's what platform is using, and focusing our energies on getting the "DELETE" value in some of these snapshots tables where they're missing, since this time window method is a temporary bandaid anyway

thomasyu888: So when platform was doing analytics, they were doing 14 days on de-duplication. (They do not de-duplicate themselves, but when they query athena, they have to de-duplicate as well)

philerooski: @jaymedina Isn't that documentation written by [Tom and yourself](https://sagebionetworks.jira.com/wiki/spaces/DW/history/3034775570/Raw+tables+in+Athena+Snowflake)? I'll ask platform whether they have it in their own documentation.

jaymedina: Yep, I assume that part of the docs came from Tom's conversations with platform but double-checking sounds good. Thanks!

thomasyu888 (approved):
🔥 LGTM!

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
```diff
@@ -0,0 +1,26 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE DYNAMIC TABLE IF NOT EXISTS ACCESSREQUIREMENT_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+    WITH latest_unique_rows AS (
```
philerooski, line 7: minor nit: I found this a bit confusing to read, and I noticed that they indent the `<query>` after the AS in the CREATE DYNAMIC TABLE [examples](https://docs.snowflake.com/en/sql-reference/sql/create-dynamic-table#examples). Do we want to adopt that formatting pattern?

philerooski (approved):

synapse_data_warehouse/synapse/dynamic_tables/V2.23.0__accessrequirement_latest.sql
jaymedina: Yep! That works for me. Updated: https://github.com/Sage-Bionetworks/snowflake/pull/84/commits/05ff362a930047f4ea043f2b1c13013d871d8138

thomasyu888 (approved):
🔥 LGTM!

PR #87, opened 2024-11-20
[SNOW-82] Initialize dynamic table: `projectsetting_latest`

jaymedina:
### problem

About the data: The [Project Setting object](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/project/ProjectSetting.html) on Synapse is a project-based setting. The `projectsettingsnapshots` table contains records of all the states of a given project setting object. This includes when it was created, who it was created by, how it was last changed, and by whom, as well as other metadata like the setting type and what project this setting is for.

The latest table for `projectsettingsnapshots` does not exist in Snowflake. We should introduce a dynamic table that will reflect the latest state of a given project setting configuration on Synapse (i.e. whether it has been created, updated, or deleted, and all metadata which reflects this change)

### solution

A new Version script is created to introduce the dynamic table `projectsetting_latest` 

### testing

There are 2,282 rows in the final table; one for the latest state of each individual project setting configuration

image

From the snapshots table, we can confirm that there are 2,282 unique IDs

image

This query ensures that all IDs are accounted for

image

And this query ensures that the LATEST state is indeed the one that is reflected in the final table

image

### DIFF
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.24.0__projectsettings_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.24.0__projectsettings_latest.sql
@@ -1,0 +1,24 @@
+use schema {{database_name}}.synapse; --noqa: JJ01,PRS,TMP,CP01
+
+CREATE DYNAMIC TABLE IF NOT EXISTS PROJECTSETTING_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+        WITH latest_unique_rows AS (
+            SELECT
+                *
+            FROM
+                {{database_name}}.synapse_raw.projectsettingsnapshots --noqa: TMP
+            WHERE
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '14 DAYS'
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM! will defer to Phil for a final review

@jaymedina one thing we may want to consider is doing some pre-processing on the array column for the "locations" but that's also not necessary.

jaymedina:
@thomasyu888 I agree we could do some pre-processing and I wonder if we'd want that to be in separate dynamic tables like `projectsetting_prepro`, which would be the cleaned up version of the `snapshots` tables, and what we would derive the latest tables from.

So something like:

```
snapshots -> prepro -> latest
```

But I think this is a larger conversation related to the medallion architecture and we could make a ticket to write up a design doc for it.

cc @philerooski

thomasyu888:
@jaymedina just adding in my 2cents. That's a good idea, but being that the latest tables are mostly the silver layer tables, that's where this kind of transform could happen.  The difference being: we transform every row (ETL) when it's appended into raw snapshot tables, or transform just the N=1 (transform what you need) (ELT)

Whether we actually do that transform also depends on how we will use the data in that column. If we won't use it as an array, it may not be worthwhile to transform.

PR #88, opened 2024-11-23
[SNOW-160] Add `version_history` column to nodesnapshot table

philerooski:
I structured the PR as three "atomic" versioned scripts:

`synapse_data_warehouse/synapse_raw/tables/V2.23.0__add_version_history_to_nodesnapshot.sql`

- Adds the `version_history` column with comment to the `nodesnapshots` table.

`synapse_data_warehouse/synapse_raw/tasks/V2.23.1__update_node_tasks.sql`

- Updates the data-loading task associated with the `nodesnapshots` table.

`synapse_data_warehouse/synapse_raw/tables/V2.23.2__refresh_node_snapshot.sql`

- Removes existing records from the `nodesnapshots` table and performs the same `COPY INTO` operation as the data-loading task with `FORCE=TRUE`. I opened a Jira ticket [here](https://sagebionetworks.jira.com/browse/SNOW-189) to streamline this process in the future.


synapse_data_warehouse/synapse_raw/tables/V2.23.0__add_version_history_to_nodesnapshot.sql
```diff
@@ -0,0 +1,9 @@
+-- Configure environment
+USE ROLE sysadmin;
+USE SCHEMA {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP,CP02
+USE WAREHOUSE compute_xsmall;
+
+-- Add `version_history` column
+ALTER TABLE nodesnapshots
```
thomasyu888, line 7: Should we just use a R script and do `CREATE OR ALTER` and take the time to document all the columns

synapse_data_warehouse/synapse_raw/tables/V2.23.2__refresh_node_snapshot.sql
```diff
@@ -0,0 +1,56 @@
+-- Configure environment
+USE ROLE sysadmin;
+USE SCHEMA {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP,CP02
+USE WAREHOUSE compute_medium;
+
+-- Drop all records from this table
+TRUNCATE TABLE nodesnapshots;
```
thomasyu888, line 7: I'm ok without truncating the table, there will be slight data loss, but we can continue copying data.  By truncating the data it will take longer than an hour to run the processing which might cause this CI/CD pipeline to timeout.

synapse_data_warehouse/synapse_raw/tasks/V2.23.1__update_node_tasks.sql
```diff
@@ -0,0 +1,63 @@
...
+USE SCHEMA {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP,CP02
+USE WAREHOUSE compute_xsmall;
+
+-- suspend nodesnapshot task and its dependents
+ALTER TASK nodesnapshot_task SUSPEND;
+ALTER TASK upsert_to_node_latest_task SUSPEND;
+ALTER TASK remove_deleted_nodes_task SUSPEND;
+
+-- Add `version_history` column to task definition
+ALTER TASK nodesnapshot_task MODIFY AS
```
thomasyu888, line 12: Should this be a repeatable script?  `CREATE OR ALTER` task

synapse_data_warehouse/synapse_raw/tables/V2.23.0__add_version_history_to_nodesnapshot.sql
philerooski: And as a general rule of thumb, changes to tables ought to be versioned. Although I'll admit that this reasoning gets a bit convoluted when you consider that we always have the git version (commit sha), so versioning changes with schemachange can seem redundant at times.

synapse_data_warehouse/synapse_raw/tables/V2.23.2__refresh_node_snapshot.sql
philerooski: @thomasyu888 In that case, should we remove this versioned script altogether? If we don't truncate/backfill, the table will forever be missing `version_history` for those snapshots which have this field. On the other hand, the practical aspect of having the `version_history` for every node will eventually be taken care of, as you said. That's assuming that this table behaves like a snapshot table and not a change capture table, a concern I brought up with Sandhra [here](https://sagebionetworks.slack.com/archives/C05PBT2FUHF/p1732647867155019?thread_ts=1732554003.064199&cid=C05PBT2FUHF).

thomasyu888: One final comment then I'm ooo.

1. There are change timestamps captures within the table when there's an action like create, update or delete.
2. Every week, there is a snapshot timestamp that captures all the actions within the week or perhaps captures whatever is the current state within the data warehouse (this we need to confirm more)


To add to further question for your question to sandrah, when new data is added, it's not always backfilled with all the variables. Which means that the data for that field will exist when there is a new change to the entity, then the snapshot will capture that on a weekly basis.

The snapshot is NOT a snapshot of what's within the database (_i think_) but rather what's in the data warehouse.

philerooski: @thomasyu888

> whether or not the weekly snapshots contain all the data that's been snapshotted previously OR just the snapshots of the actions for that week.

I queried the production data. I found approximately the same number of unique, non-deleted nodes in the `nodesnapshots` table as the number of nodes which were snapshotted in the last 8 days. I also found nodes which haven't been changed recently yet were included in the recent snapshot. So my thinking is that _every_ existing (i.e. `change_type` != 'DELETE') node, regardless of when it was last changed, is included in the weekly snapshot.

What this means in the context of this PR is that the new `version_history` column will be populated for every existing node in the next weekly snapshot without us needing to truncate and re-`COPY INTO` the `nodesnapshots` table. But any nodes which were deleted before the next snapshot and since `version_history` began being collected won't have this data "backfilled" automatically. If we don't care about including version history metadata for such nodes (~3 million nodes, or ~5% of the unique nodes in the `nodesnapshots` table), then we should get rid of this versioned script.

philerooski: Alternatively, if we do want to preserve this metadata but we don't want to re-populate the entire `nodesnapshots` table, we might be able to `DELETE` any records which have a snapshot date in the last ~3 weeks and then force `COPY INTO` just those records from the stage. That's ~90 million records, or ~2% of the total records in the `nodesnapshots` table.

thomasyu888: @philerooski 

What is your recommendation based off of your findings?  it sounds like what we would be missing is `version_snapshot` for the `DELETE` records?  If that is the case, those would be removed anyhow from the latest tables.

philerooski: @thomasyu888

I don't like the idea of throwing away historical data because it won't impact the most up-to-date data. 90 million records is not such a huge amount of data to copy into Snowflake. I recommend deleting and copying data selectively so that our `nodesnapshots` table in Snowflake is up-to-date with the data warehouse.

thomasyu888: @philerooski 

I like that idea a lot! Let's go with your recommendation.  Thanks for looking into the data!  For the userprofile snapshots, one of the columns haven't been backfilled in awhile, so for that snapshot, it may actually make sense to just do a whole sail replacement, but I would also look at the production data first.

BryanFauble (approved):
LGTM!

synapse_data_warehouse/synapse_raw/tables/V2.23.2__refresh_node_snapshot.sql
philerooski: @thomasyu888 

What's annoying about doing this in practice is that there's no perfect way to say `COPY INTO` all files from this date onward. The only filtering we can do as part of the `COPY INTO` statement is regex. Theoretically, I could invent some horrendous pattern which says something like "all files from 2024-11-* and 2024-12-* and all files for the next 100 years after that.

Instead, we might be tempted to query the last modified date of a file by `LIST`ing a stage and `COPY INTO`ing only those files, but Snowflake gives this to us as an ugly, non-ISO string like "Wed, 2 Oct 2024 00:02:40 GMT" that can't be easily cast as a datetime. Probably we would rather parse out the `snapshot_date=*` component of the S3 prefix and cast that instead.

The bigger picture problem here is that when including `COPY INTO` statements in versioned scripts, the end state of the target table is dependent on the data in the stage -- which is constantly changing. So if there was ever a need to roll back to a previous version of the database and replay these changes we wouldn't get the same result.

/rant I'm going to go implement the horrendous regex now

philerooski:
@thomasyu888 @BryanFauble 

My most recent commit is a bit convoluted. There's some explanation in the SQL comments for posterity, but to expand further: backfilling `version_history` for data which has this field in the stage data but not in the actual `nodesnapshots` table (since the data was loaded before we added the `version_history` column) entailed avoiding a few obstacles:

1. `nodesnapshots` is too big to simply `TRUNCATE` the table and reload all stage data. We can strategically `DELETE` rows which may have `version_history` (it was added to the data warehouse on 2024-11-12), but strategically refreshing those same rows is more complicated. `COPY INTO` only supports filtering files on a regex, which is what I do here. I talk more about why this is so difficult [here](https://github.com/Sage-Bionetworks/snowflake/pull/88#discussion_r1868516209).
2. There's some, let's say... semantic complications in using `COPY INTO` in a versioned script. While the imperative sequence of steps remains the same, the data in our stage is always changing, as is the load history (`INFORMATION_SCHEMA.COPY_HISTORY`) of our table. I made a blunt attempt at avoiding a `COPY INTO` operation in the off-chance that this script is being executed as part of a rollback/replay deployment by first checking whether there are even any snapshots in our table which could have `version_history` (`snapshot_date >= '2024-11-01'`).

synapse_data_warehouse/synapse_raw/tables/V2.23.2__refresh_node_snapshot.sql
```diff
@@ -3,54 +3,77 @@ USE ROLE sysadmin;
...
+            PARSE_JSON($1:scope_ids) AS scope_ids,
+            PARSE_JSON($1:items) AS items,
+            PARSE_JSON($1:reference) AS reference,
+            $1:is_search_enabled AS is_search_enabled,
+            $1:defining_sql AS defining_sql,
+            PARSE_JSON(REPLACE(REPLACE($1:internal_annotations, '\n', '\\n'), '\r', '\\r')) AS internal_annotations,
+            PARSE_JSON(REPLACE(REPLACE($1:version_history, '\n', '\\n'), '\r', '\\r')) AS version_history
+        FROM @{{stage_storage_integration}}_stage/nodesnapshots/ --noqa: TMP
+    )
+    PATTERN = '.*nodesnapshots/snapshot_date=(2024-(11|12)-.*|202[5-9]-.*|20[3-9]-.*)/.*'
```
BryanFauble, line 75: 👋 From 76 years in the past if this ever lives that long.

BryanFauble (approved):

synapse_data_warehouse/synapse_raw/tables/V2.23.0__add_version_history_to_nodesnapshot.sql
```diff
@@ -0,0 +1,9 @@
...
+  ADD COLUMN version_history VARIANT
+  COMMENT 'The list of entity versions, at the time of the snapshot.';
```
jaymedina, line 9: nit (feel free to ignore): The list of all _previous_ entity versions

jaymedina (changes requested):
Just a heads up, there are already version scripts notated with `V2.23.0` and `V2.24.0` in DEV so the CI build will fail upon merge

synapse_data_warehouse/synapse_raw/tables/V2.23.0__add_version_history_to_nodesnapshot.sql
```diff
@@ -0,0 +1,9 @@
+-- Configure environment
+USE ROLE sysadmin;
```
jaymedina, line 2: These role and warehouse specifications are redundant since they are specified in the [CI job](https://github.com/Sage-Bionetworks/snowflake/blob/6613b12340f96dd16832da387690bdcccf8c5049/.github/workflows/ci.yaml#L60).

I'm also not a fan of making these specifications in the scripts themselves since this opens up potential inconsistencies from script to script. I think it would make more sense for the job running `schemachange` to be the final source of truth for these configurations. What do you think?

synapse_data_warehouse/synapse_raw/tables/V2.23.0__add_version_history_to_nodesnapshot.sql
philerooski: Actually, I don't think that's accurate. I see a `version_history` value that appears to include the current latest version. It looks like this:

```
[
  {
    "id": "syn9606778",
    "isLatestVersion": true,
    "modifiedBy": "1",
    "modifiedByPrincipalId": "1",
    "modifiedOn": 1731441964134,
    "versionComment": "in progress",
    "versionLabel": "in progress",
    "versionNumber": 2
  },
  {
    "id": "syn9606778",
    "isLatestVersion": false,
    "modifiedBy": "1",
    "modifiedByPrincipalId": "1",
    "modifiedOn": 1731441964101,
    "versionLabel": "1",
    "versionNumber": 1
  }
]
```

philerooski:
> Just a heads up, there are already version scripts notated with `V2.23.0` and `V2.24.0` in DEV so the CI build will fail upon merge

Thanks for the heads up! Probably we should have a check which blocks merging if we haven't rebased on dev, although that doesn't perfectly solve the problem of one developer's version stepping on the toes of another developer.

synapse_data_warehouse/synapse_raw/tables/V2.23.0__add_version_history_to_nodesnapshot.sql
philerooski: I agree. The role should be handled by the deployer, rather than the script itself. If we need to specify _who_ has OWNERSHIP of _what_ that should be handled explicitly in the script.

synapse_data_warehouse/synapse_raw/tables/V2.23.0__add_version_history_to_nodesnapshot.sql
jaymedina: oh I see - thanks for checking!

synapse_data_warehouse/synapse_raw/tables/V2.23.2__refresh_node_snapshot.sql
```diff
@@ -3,54 +3,77 @@ USE ROLE sysadmin;
 USE SCHEMA {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP,CP02
 USE WAREHOUSE compute_medium;
 
--- Drop all records from this table
-TRUNCATE TABLE nodesnapshots;
+SET num_records_to_be_reloaded = (
+    SELECT count(*)
+    FROM nodesnapshots
+    WHERE snapshot_date >= '2024-11-01'
+);
```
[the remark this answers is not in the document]
BryanFauble: From the script I saw they paginate over the 50k versions and get the most recent 50k versions.

It might be worthwhile to confirm what the behavior is here. This is the line of code I was referring to:

https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/ea0fa386da6dda0a3760b097b92883facc1c2795/services/workers/src/main/java/org/sagebionetworks/snapshot/workers/writers/NodeObjectRecordWriter.java#L120

thomasyu888 (approved):
🔥 LGTM!  Please do address any remaining comments, but thanks for implementing the complex regex and neat truncation.

jaymedina (approved):
Thanks for this work 🔥

synapse_data_warehouse/synapse_raw/tables/V2.23.2__refresh_node_snapshot.sql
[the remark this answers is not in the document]
philerooski: @BryanFauble This feels more like a potential upstream data warehouse issue. It doesn't make sense to me at this point for us to try to preserve history which our upstream process is intentionally not collecting. And, you're right, this is an extreme edge case that would take a confluence of unlikely factors to run into, or even have an impact on downstream analysis.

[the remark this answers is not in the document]
thomasyu888: The max version number is 36333, so perhaps this is a case where we "worry about it when we get there".  

Feel free to merge if you're comfortable with this PR @philerooski !  Thanks!

PR #89, opened 2024-12-05
[SNOW-175] Convert `node_latest` table -> dynamic table

jaymedina:
## problem

Currently the `node_latest` table is a regular table that gets updated using tasks and streams. We can simplify our data warehouse by turning this into a dynamic table that updates itself without requiring tasks & streams. 

## solution

- [x] Add a script to create a backup of the original `node_latest` table in case of issues
- [x] Add a script to drop the original `node_latest` table so we can create a dynamic table with the same name
- [x] Create `node_latest` dynamic table that introduces some post-processing of the RAW data
- [x] Add a script to delete the scheduled task + stream that were for the original `node_latest` table

### 1. Comparing the results between the new and old `node_latest` tables

I found a discrepancy between the results from the new `node_latest` dynamic table query and the current `node_latest` table. There are 1,413,032 rows missing from the **new** `node_latest` dynamic table:

image

As suspected, the discrepancies are due to the added time window of 14 days in the dynamic table query:

✅ 
image

### 2. Confirm there are no duplicate `id`s

✅ 
image

### 3. Confirm that all non-deleted nodes are accounted for

For this test I created a temp table that shows the IDs missing from the latest table that exist in the snapshots table. From visual inspection, it looks like these IDs are missing either because their `change_type` is set to DELETE, or because their last snapshot was taken outside of the 2-week window, in which case they are misrepresentations of the current state of the nodes.

✅ 
image

### 4. Confirm that the LATEST version of each node is represented in the new `node_latest`

✅ 
image


synapse_data_warehouse/synapse/dynamic_tables/V2.26.2__node_latest.sql
```diff
@@ -0,0 +1,36 @@
...
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+        WITH latest_unique_rows AS (
+            SELECT
+                *
+            FROM
+                {{database_name}}.synapse_raw.nodesnapshots --noqa: TMP
+            WHERE
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '14 DAYS'
```
jaymedina, line 13: This is the major difference between the `UPSERT_TO_NODE_LATEST_TASK` (log onto SYSADMIN to view this task on Snowsight) and my query. I found that not adding the 14 day window restriction slows down the query by _a lot_, so even though we can remove deleted records using `change_type`, leaving this `SNAPSHOT_TIMESTAMP` window will help with efficiency.

synapse_data_warehouse/synapse/dynamic_tables/V2.26.2__node_latest.sql
```diff
@@ -0,0 +1,36 @@
...
+                    ORDER BY change_timestamp DESC, snapshot_timestamp DESC
+                ) = 1
+        ),
+        processed_latest_unique_rows AS (
+            SELECT
+                * EXCLUDE (annotations, reference),
+                PARSE_JSON(annotations):annotations::OBJECT AS annotations,
+                PARSE_JSON(annotations):etag::STRING AS etag,
+                PARSE_JSON(reference):targetId::STRING as reference_target_id,
+                PARSE_JSON(reference):targetVersionNumber::STRING as reference_target_version_number
```
jaymedina, line 25: These last 4 lines are doing some data processing to refine `node_latest` and make it more silver 🥈 

* I'm extracting the "annotations" and "etag" keys from the JSON stored in the original `annotations` column, and putting them in their own columns. I'm also excluding the original `annotations` column from the final table.
* I'm doing the same extraction + exclusion for the `reference` column.

thomasyu888: I like this update, we may have to worry about the breaking change this makes. Many people are using annotations:annotations already so this would potentially be an issue

jaymedina: I see - since this is a breaking change and it's December, it'd make sense to be more intentional about this: 

I can remove these changes from this PR, and pitch this type of transformation in our medallion architecture design doc (do we have a SPIKE ticket for the medallion work yet?).

In that doc, we can also discuss how we would share data structure updates to users who already have dashboards/Streamlit apps deployed. I like the pseudo release notes you've been sharing on the `synapse_datawarehouse` channel, so we could expand on that so it reaches all users.

LMK what you guys think cc @philerooski

thomasyu888: @jaymedina 

Just my 2cents, but I think we can create a ticket around creating our first set of dashboard-ready tables.  For example, extracting annotations _could_ be a "gold" table.

Non-backwards compatible changes would be handled like other products - I can float to our stakeholders this proposed change in the future and ask them if that is desired.  From there, we would have a timeline for implementation and migration plans

thomasyu888 (commented):
🔥 Awesome work here - I'll leave it to Phil for a final review. I'm excited about the transition to dynamic tables!

philerooski (approved):

thomasyu888:
@jaymedina / @philerooski I wouldn't merge this just yet, because if it is merged then the highest V scripts would be recognized as 2.26 and skip over the work you're doing Phil.

synapse_data_warehouse/synapse_raw/tasks/V2.26.1__delete_node_latest_tasks.sql
```diff
@@ -0,0 +1,4 @@
+USE SCHEMA {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP
+
+DROP TASK IF EXISTS UPSERT_TO_NODE_LATEST_TASK;
```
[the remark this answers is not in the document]
jaymedina: > Can you delete tasks that are unsuspended?

Ooo good point. From the docs it sounds like the root task _does_ need to be suspended before any dependent task can be dropped ([Usage Notes](https://docs.snowflake.com/en/sql-reference/sql/drop-task#usage-notes)).

This quote [here](https://docs.snowflake.com/en/sql-reference/sql/alter-task#usage-notes) says that any suspended task will execute its current run to completion, so we won't be interrupting anything by suspending the root momentarily: "_When a task is suspended, any current run of the task (i.e. a run with an EXECUTING state in the [TASK_HISTORY](https://docs.snowflake.com/en/sql-reference/functions/task_history) output) is completed._".

And then it looks like we can resume the root task and all of its dependents with a [simple query](https://docs.snowflake.com/en/sql-reference/functions/system_task_dependents_enable#syntax).

I'll update this script accordingly. Thanks for pointing this out!

> May want to change the order of these task deletions, as the delete task depends on the upsert task.

I don't think the order matters here - It sounds like child tasks just become standalones and auto-suspend ([source](https://docs.snowflake.com/en/sql-reference/sql/drop-task#usage-notes)):

_"If a predecessor task in a task graph is dropped, then all former child tasks that identified this task as the predecessor become either standalone tasks or root tasks, depending on whether other tasks identify these former child tasks as their predecessor. These former child tasks are suspended by default and must be resumed manually."_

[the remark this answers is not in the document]
thomasyu888: Thanks for looking into that. There is a convenience function to restart all the tasks as well, you can look at Phil's most recent PR.

jaymedina:
> For testing, it's a bit nuanced

I'll add this to our CI/CD design doc so we can explore potential test cases.

philerooski (approved):

PR #92, opened 2024-12-10
[SNOW-160] Remove scripting block and reduce scope of refresh

philerooski:
I set up schemachange locally and ensured that by removing the scripting block I was able to successfully deploy the changes here (on a clone).

To limit the amount of data that needs to be loaded in the unlikely event we need to rollback/replay these changes, I changed the data being deleted/reloaded from all data from November 2024 onward to only data from November and December 2024. The side effect here is that any data from January 2025 onward which has been loaded into the table before we've replayed the `V2.25.*` scripts won't have its `version_history` data backfilled.


thomasyu888 (approved):
🔥 LGTM!  It'll probably be pretty unlikely that we will replay these events. 

That is one reason why I constantly struggle with having R scripts to store constant state of objects vs these V scripts that keep a history of actions that we did on objects.

PR #93, opened 2024-12-13
[SNOW-175] `node_latest` dynamic table query update: Add stricter filter to remove deleted nodes

jaymedina:
## problem

It was found that some nodes were incorrectly passing through the `CHANGE_TYPE != 'DELETE'` filter in the original query due to a misrepresentation of the latest `CHANGE_TYPE` for some deleted entities.

The reason for this misrepresentation is unclear, but is indirectly due to entities sitting in the trash can for 30 days leading to them being purged without a `DELETE` being recorded for this event. It also appears that this has something to do with the `PARENT_ID` of a node being changed, since the example we've seen this for had all their `CHANGE_TIMESTAMP`s locked on the date in which the parent entity was changed. As a result of this, all `CHANGE_TYPE`s were recorded with the same timestamp down to the second, resulting in `UPDATE` being the latest `CHANGE_TYPE` for these already-deleted nodes and leading to ghost nodes in the final latest table.

image

## solution

A [Jira ticket](https://sagebionetworks.jira.com/browse/PLFM-8747?atlOrigin=eyJpIjoiZTkzYjU4MmIyMWZhNDcxYmE5MTAxNjc0NTIxZjNkNTUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) was created for the Platform team, and in this PR we introduce a workaround in the meantime, that involves stricter filtering of nodes that are deleted on Synapse. The trash can on Synapse behaves as a Synapse entity, therefore it has its own Synapse ID: `1681355`.

If we update the query to filter out nodes with a parent or benefactor of that same ID, we should be able to catch all deleted nodes.

## testing

The new solution is still not capturing all the deleted nodes, but we are getting better results than before.

The discrepancies that remain are a mix of entities that still exist but for some reason they haven't been snapshotted, and entities that were deleted but it's not represented on Snowflake. Entity `syn64129904` is an example of one that was deleted but its deletion was never recorded on Snowflake. It had no recorded change to its parent entity.

Results from original solution:
For 14 days - 25,837,058 rows
For 30 days - 25,863,689 rows

Results from new solution
For 14 days - 25647157
For 30 days - 25647216

In other words, the num. of discrepancies were brought down from 26,631 to 59 (as of Dec 13, 2024)

### DIFF
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.26.5__remove_ghost_assets_node_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.26.5__remove_ghost_assets_node_latest.sql
@@ -10,7 +10,7 @@ CREATE OR REPLACE DYNAMIC TABLE NODE_LATEST
             FROM
                 {{database_name}}.synapse_raw.nodesnapshots --noqa: TMP
             WHERE
-                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '30 DAYS'
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '21 DAYS'
             QUALIFY ROW_NUMBER() OVER (
                     PARTITION BY id
                     ORDER BY change_timestamp DESC, snapshot_timestamp DESC
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.26.5__remove_ghost_assets_node_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.26.5__remove_ghost_assets_node_latest.sql
@@ -10,7 +10,7 @@ CREATE OR REPLACE DYNAMIC TABLE NODE_LATEST
             FROM
                 {{database_name}}.synapse_raw.nodesnapshots --noqa: TMP
             WHERE
-                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '21 DAYS'
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '30 DAYS'
             QUALIFY ROW_NUMBER() OVER (
                     PARTITION BY id
                     ORDER BY change_timestamp DESC, snapshot_timestamp DESC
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM!

thomasyu888 (approved):
🔥 LGTM!  Will leave it to @philerooski for final review.

synapse_data_warehouse/synapse/dynamic_tables/V2.26.5__remove_ghost_assets_node_latest.sql
```diff
@@ -0,0 +1,26 @@
...
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+        WITH latest_unique_rows AS (
+            SELECT
+                *
+            FROM
+                {{database_name}}.synapse_raw.nodesnapshots --noqa: TMP
+            WHERE
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '30 DAYS'
```
philerooski, line 13: Do we need to increase the time window from 14 to 30 days? Looking at the node you provided as an example in PLFM-8747:

image

While the node is in the trash can it's still being snapshotted as part of the weekly snapshots. So it's sufficient to look at the regular 14-day window and filter where benefactor/parent is the trash can.

jaymedina: There are some nodes that still exist on Synapse but have snapshots that are outdated by ~4 weeks (for example [26262846](https://www.synapse.org/Synapse:syn26262846)) and so they're only captured by the query when the window is 30 days:

image

I haven't yet explored growing the window beyond 30 days, to 40+. I'll run some queries to see if we are missing enough entities that it makes expanding the window worth it (right now the query runs for 6+ mins when the window is 30 days)

philerooski: I see, so this PR is not just trying to address the deleted nodes issue, but also mitigate this issue:
> entities that still exist but for some reason they haven't been snapshotted, and entities that were deleted but it's not represented on Snowflake 

6+ minutes is not too intensive, so I think the current implementation is fine. Although, considering that this affects just ~0.1% of nodes (if I understand your "testing" section correctly) and we can expect an upstream fix (or, at least, find out the true scope/cause of the issue) from platform, we don't need to be too fussy about the time window at this point.

philerooski (approved):

synapse_data_warehouse/synapse/dynamic_tables/V2.26.5__remove_ghost_assets_node_latest.sql
jaymedina: Sounds good to me- I ran some more tests and expanded the window up to 60 days. We get a dozen or so new rows every time and yes most of the time these nodes have already been deleted. Let's go with the 30 day window for now and we can iterate on this based on what we hear back from the platform team.

PR #95, opened 2024-12-18
[SNOW-60,162,165] Add new columns to `userprofilesnapshots`

jaymedina:
## problem

The parquet files now contain a new pieces of metadata `is_two_factor_auth_enabled` `industry` and `tos_agreements` but this metadata is not reflected in the current `userprofilesnapshots` table.

## solution

Following this SOP:
- [x] Write a V script to add the new columns to the snapshots table
- [x] Write a V script to alter the existing task that copies new data into the snapshots table
- [x] Write a V script to manually refresh the entire snapshots table with the new column values

## testing

I tested the queries of all the scripts in `SYNAPSE_DATA_WAREHOUSE_JMEDINA.synapse_raw.userprofilesnapshot` so the results are available there.

I did a cross-check using the `id` column for the `userprofilesnapshot` in `synapse_data_warehouse_dev` and `synapse_data_warehouse_jmedina` and can confirm that all the IDs are accounted for:

image

The number of rows are also the same:

image

Let me know if you'd like me to share the worksheet where I ran these tests

### DIFF
--- a/synapse_data_warehouse/synapse_raw/tables/V2.28.1__update_userprofilesnapshots_task.sql
+++ b/synapse_data_warehouse/synapse_raw/tables/V2.28.1__update_userprofilesnapshots_task.sql
@@ -25,9 +25,6 @@ ALTER TASK userprofilesnapshot_task MODIFY AS
             $1:location as location,
             $1:company as company,
             $1:position as position,
-            $1:is_two_factor_auth_enabled as is_two_factor_auth_enabled,
-            $1:industry as industry,
-            $1:tos_agreements as tos_agreements,
             NULLIF(
                 REGEXP_REPLACE(
                     metadata$filename,
@@ -35,7 +32,10 @@ ALTER TASK userprofilesnapshot_task MODIFY AS
                 ),
                 '__HIVE_DEFAULT_PARTITION__'
             ) as snapshot_date,
-            $1:created_on as created_on
+            $1:created_on as created_on,
+            $1:is_two_factor_auth_enabled as is_two_factor_auth_enabled,
+            $1:industry as industry,
+            $1:tos_agreements as tos_agreements
         from
             @synapse_prod_warehouse_s3_stage/userprofilesnapshots --noqa: TMP
     )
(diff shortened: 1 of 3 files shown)


synapse_data_warehouse/synapse_raw/tables/V2.28.1__update_userprofilesnapshots_task.sql
```diff
@@ -0,0 +1,45 @@
...
+            $1:user_name as user_name,
+            $1:first_name as first_name,
+            $1:last_name as last_name,
+            $1:email as email,
+            $1:location as location,
+            $1:company as company,
+            $1:position as position,
+            $1:is_two_factor_auth_enabled as is_two_factor_auth_enabled,
+            $1:industry as industry,
+            $1:tos_agreements as tos_agreements,
```
philerooski, line 30: When I was updating the node snapshots table with a VARIANT `version_history`, I had the task load the column like this:

```
PARSE_JSON(...) AS version_history
```

in order to convert the JSON string to a VARIANT and so that it would be consistent with how other VARIANT columns are handled in the existing task. But apparently, if our stage data is valid JSON, by loading `tos_agreements` into a VARIANT column as you have Snowflake will implicitly convert the JSON string to a VARIANT. So the way you've done it here is cleaner.

If we look at the full command I was using (which I copied from other VARIANT columns in the same task):
```
PARSE_JSON(REPLACE(REPLACE($1:version_history, '\n', '\\n'), '\r', '\\r')) AS version_history
```
it must have been the case, at least in the past, that some of the JSON were pretty-printed. You might want to clone the production data and verify that loading `tos_agreements` implicitly as you have doesn't run into any pretty-printed JSON issues.

philerooski (approved):

philerooski:
Whoops I think I meant to leave a "comment" review. Please re-request my review once you're finished.

synapse_data_warehouse/synapse_raw/tables/V2.28.1__update_userprofilesnapshots_task.sql
jaymedina: Thanks for this @philerooski! The issue ended up being that the order in which I was introducing the columns in scripts 2 and 3 did not match the table structure. In script 1, adding the new columns just tacks them on at the end. In scripts 2 and 3 I have the new columns ordered from 2nd to last, so the wrong metadata was going into the wrong column! I'll make a note of this in the SOP and update the scripts in my next commit.

image

There is another issue I need to sort out which is the empty `industry` values, but everything else looks good (row numbers are the same, etc), so I'll mark this as RFR and tag you again after that

jaymedina: It works! I tested this with Tom's account since he has this field filled out. This is ready for review now

image

jaymedina: image

philerooski (approved):

PR #96, opened 2024-12-20
[SNOW-60] Hotfix: Programmatically feed stage name

jaymedina:
Instead of programmatically feeding the stage name, I had it hardcoded to `prod`, which broke when deploying to `dev`.

This PR patches that up in both the task and refresh scripts.

### DIFF
--- a/synapse_data_warehouse/synapse_raw/tables/V2.28.1__update_userprofilesnapshots_task.sql
+++ b/synapse_data_warehouse/synapse_raw/tables/V2.28.1__update_userprofilesnapshots_task.sql
@@ -37,7 +37,7 @@ ALTER TASK userprofilesnapshot_task MODIFY AS
             $1:industry as industry,
             $1:tos_agreements as tos_agreements
         from
-            @{{stage_storage_integration}}/userprofilesnapshots --noqa: TMP
+            @{{stage_storage_integration}}_stage/userprofilesnapshots --noqa: TMP
     )
     pattern = '.*userprofilesnapshots/snapshot_date=.*/.*';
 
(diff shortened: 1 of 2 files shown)


philerooski (approved):

synapse_data_warehouse/synapse_raw/tables/V2.28.1__update_userprofilesnapshots_task.sql
```diff
@@ -37,7 +37,7 @@ ALTER TASK userprofilesnapshot_task MODIFY AS
             $1:industry as industry,
             $1:tos_agreements as tos_agreements
         from
-            @synapse_prod_warehouse_s3_stage/userprofilesnapshots --noqa: TMP
+            @{{stage_storage_integration}}/userprofilesnapshots --noqa: TMP
```
thomasyu888, line 40: Do not merge: 

```
        FROM @{{stage_storage_integration}}_stage/nodesnapshots/ --noqa: TMP
```

thomasyu888 (commented):
🔥 LGTM! Sorry about the nuance around the stage name env variable. I defer to @philerooski for final re-review.

philerooski (approved):
Huh I thought I already approved the hotfix.

jaymedina:
I had to add the `_stage` part

philerooski:
@jaymedina ah, okay. For very minor changes, if I've already approved then there's no need to re-request. I don't want to be a blocker for typo fixes etc.

thomasyu888:
@philerooski, for transparency, I had requested for your final re-review, as I noticed that typo and made a comment.  Another thing I noticed was that the COPY into script isn't using the COMPUTE_MEDIUM warehouse, so copying into for the production warehouse is going to take that much longer if it took 15 minutes to copy ~400k rows.

jaymedina:
> Another thing I noticed was that the COPY into script isn't using the COMPUTE_MEDIUM warehouse, so copying into for the production warehouse is going to take that much longer if it took 15 minutes to copy ~400k rows.

I see - then I think it's worth establishing a set of guidelines somewhere for what warehouse to use for a given script (e.g. if you are loading >500k rows worth of data, use MEDIUM)

We can use the docs [here](https://docs.snowflake.com/en/user-guide/warehouses-overview#impact-on-data-loading) for reference

In the meantime, I'll make another PR to update to MEDIUM

PR #97, opened 2024-12-20
Update nodesnapshots using medium warehouse

philerooski:
From @thomasyu888 https://github.com/Sage-Bionetworks/snowflake/pull/96#issuecomment-2557772001

### DIFF
--- a/synapse_data_warehouse/synapse_raw/tasks/V2.25.1__update_node_tasks.sql
+++ b/synapse_data_warehouse/synapse_raw/tasks/V2.25.1__update_node_tasks.sql
@@ -1,6 +1,6 @@
 -- Configure environment
 USE SCHEMA {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP,CP02
-USE WAREHOUSE compute_xsmall;
+USE WAREHOUSE compute_medium;
 
 -- suspend nodesnapshot task and its dependents
 ALTER TASK refresh_synapse_warehouse_s3_stage_task SUSPEND;


synapse_data_warehouse/synapse_raw/tasks/V2.25.1__update_node_tasks.sql
```diff
@@ -1,6 +1,6 @@
 -- Configure environment
 USE SCHEMA {{database_name}}.synapse_raw; --noqa: JJ01,PRS,TMP,CP02
-USE WAREHOUSE compute_xsmall;
```
thomasyu888, line 3: This isn't the script that requires compute medium. Let's chat more on monday

philerooski: Ah, okay. I thought maybe it needed it since we are probably loading over 100 million records.

thomasyu888: You have that here: https://github.com/Sage-Bionetworks/snowflake/blob/dev/synapse_data_warehouse%2Fsynapse_raw%2Ftables%2FV2.25.2__refresh_node_snapshot.sql#L2-L4

philerooski: I need to stop working 😓

thomasyu888: 😅 you and me both. Almost the holidays.

PR #98, opened 2024-12-20
[SNOW-60] Update the warehouse for the large refresh

jaymedina:
The refresh takes >10 mins to run for 400K rows in `dev` so it's worth bumping it up from xsmall->medium for the large refresh, since `prod` has >14m rows

The task will use xsmall since that's an incremental update


thomasyu888 (approved):
🔥 LGTM! It's ok that this won't run on the dev pipeline again because the dev pipeline has ingested the data.

PR #99, opened 2025-01-08
[SNOW-171] Convert `certifiedquizquestion_latest` table -> dynamic table

jaymedina:
## problem

Currently this latest table is a regular table that gets updated using tasks and streams. We can simplify our data warehouse by turning this into a dynamic table that updates itself without requiring tasks & streams.

## solution

Following the instructions of [this SOP](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3794894849/WIP+Converting+tables+to+dynamic+tables):

- [x] Add a script to create a backup of the original table in case of issues
- [x] Add a script to drop the task + stream for the original table
- [x] Add a script to drop the original table so we can create a dynamic table with the same name
- [x] Create dynamic table that introduces some post-processing of the RAW data
- [x] Add table and column comments to dynamic table
- [x] Add a script to delete the scheduled task + stream that were for the original table

## testing

* [Link to cloned DB where all scripts were run](https://app.snowflake.com/mqzfhld/vp00034/#/data/databases/SYNAPSE_DATA_WAREHOUSE_JMEDINA/schemas/SYNAPSE/dynamic-table/CERTIFIEDQUIZQUESTION_LATEST/data-preview)

* Data validation:

1. Ensure there are no duplicates in the latest table

image

2. Ensure no missing questions for each `response_id` in the latest table compared to snapshot table

```
with snapshot_ids as (
    select
        response_id,
        question_index,
        max(snapshot_timestamp) as latest_snapshot_timestamp,
        max(change_timestamp) as latest_change_timestamp
    from 
        temp_latest_unique_rows
    group by 
        response_id, question_index
),
-- 2. Then we get all the unique IDs from latest
latest_ids as (
    select 
        response_id, question_index
    from 
        temp_latest_unique_rows
),
-- 3. Identify missing IDs and exclude those present in both
missing_ids as (
    SELECT 
        COALESCE(snapshot_ids.response_id, latest_ids.response_id) AS response_id,
        COALESCE(snapshot_ids.question_index, latest_ids.question_index) AS question_index,
        CASE 
            WHEN snapshot_ids.response_id IS NULL THEN 'missing_from_snapshots'
            WHEN latest_ids.response_id IS NULL THEN 'missing_from_latest'
            WHEN snapshot_ids.question_index IS NULL THEN 'question_index_missing_from_snapshots'
            WHEN latest_ids.question_index IS NULL THEN 'question_index_missing_from_latest'
        END AS missing_status,
[...]

PR #100, opened 2025-01-08
[SNOW-155] Converting userprofile_latest table to dynamic table

danlu1:
Problem:

Currently, the userprofile_latest table is a table that is updated via upserting task that is based on userprofilesnapshot stream. We want to convert it to dynamic table so it gets refreshed automatically. 

Solution:

- [x] Add a script to create a backup of the original userprofile_latest table which need to be removed manually after validation
- [x] Add a script to delete the upserting task for the original userprofile_latest table
- [x] Add a script to delete stream for userprofilesnapshot table that is referred in userprofile_latest upserting task
- [x] Add a script to drop the original userprofile_latest table so we can create a dynamic table with the same name
- [x] Create userprofile_latest dynamic table and convert empty strings in `LOCATION`, `COMPANY`, `POSITION`, `INDUSTRY` to Null for easy search. Keep `TOS_AGREEMENTS` as VARIANT variable as Phil's comments.
- [x] Add a script to add table and column comments.
 
Validation(edited):
As the instructions in [Validating a "latest" table on synapse_data_warehouse_dev](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3809738755/Validating+a+latest+table+on+synapse_data_warehouse_dev):
1. Ensure there are no duplicates in the latest table
```
select
    count(*) as total_rows,
    count(distinct id) as total_unique_ids
from SYNAPSE_DATA_WAREHOUSE_DANLU_STAGE.SYNAPSE.USERPROFILE_LATEST_test2;
```
Screenshot 2025-01-08 at 5 12 49 PM

2. Ensure no missing IDs be in latest table compared to snapshot table 
```
with snapshot_ids as (
    select
        id,
        max(snapshot_timestamp) as latest_snapshot_timestamp,
        max(change_timestamp) as latest_change_timestamp
    from 
        SYNAPSE_DATA_WAREHOUSE_DANLU_STAGE.SYNAPSE_RAW.userprofilesnapshot
    group by 
        id
),
-- 2. Then we get all the unique IDs from latest
latest_ids as (
    select 
        id
    from 
        SYNAPSE_DATA_WAREHOUSE_DANLU_STAGE.SYNAPSE.USERPROFILE_LATEST_test2
),
-- 3. Identify missing IDs and exclude those present in both
missing_ids as (
    select 
        coalesce(snapshot_ids.id, latest_ids.id) as id,
        case 
            when snapshot_ids.id is null then 'missing_from_snapshots'
            when latest_ids.id is null then 'missing_from_latest'
        end as missing_status,
        snapshot_ids.latest_snapshot_timestamp,
        snapshot_ids.latest_change_timestamp
    from 
        snapshot_ids
    full outer join 
        latest_ids
    on 
        snapshot_ids.id = latest_ids.id
    where 
        snapshot_ids.id is null 
        or 
        latest_ids.id is null
)
select * from missing_ids
where latest_snapshot_timestamp > current_timestamp - INTERVAL '14 days';
```
Screenshot 2025-01-08 at 5 14 10 PM

3. Ensure the empty string replacement works as expected

**Before**: 
Screenshot 2025-01-07 at 8 51 59 PM

**After**:
Screenshot 2025-01-07 at 8 57 34 PM

Depends on #99


synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
```diff
@@ -0,0 +1,49 @@
...
+	LAST_NAME VARCHAR(16777216) COMMENT 'The last name of the user.',
+	EMAIL VARCHAR(16777216) COMMENT 'The primary email of the user.',
+	SNAPSHOT_DATE DATE COMMENT 'The data is partitioned for fast and cost effective queries. The snapshot_timestamp field is converted into a date and stored in the snapshot_date field for partitioning. The date should be used as a condition (WHERE CLAUSE) in the queries.',
+	CREATED_ON TIMESTAMP_NTZ(9) COMMENT 'The creation time of the user profile.',
+	IS_TWO_FACTOR_AUTH_ENABLED BOOLEAN COMMENT 'Indicates if the user had two factor authentication enabled when the snapshot was captured.',
+    TOS_AGREEMENTS VARIANT COMMENT 'Contains the list of all the term of service that the user agreed to, with their agreed on date and version.',
+    LOCATION VARCHAR(16777216) COMMENT 'The location of the user.',
+	COMPANY VARCHAR(16777216) COMMENT 'The company where the user works.',
+	POSITION VARCHAR(16777216) COMMENT 'The position of the user in the company.',
+	INDUSTRY VARCHAR(16777216) COMMENT 'The industry/discipline that this person is associated with.'
```
jaymedina, line 21: I wonder if explicitly listing out the columns like this would ever clash with the `SELECT *` from the main query, if say `userprofilesnapshot` gets a column added to it later down the line? So either we keep the `SELECT *` and implicitly call the columns in this script + introduce the column comments in a separate script, or the columns selected are explicitly written out in the main query of this script

synapse_data_warehouse/synapse_raw/tasks/V2.29.1__delete_userprofile_latest_tasks.sql
```diff
@@ -0,0 +1,8 @@
+-- Drop any scheduled tasks
+USE SCHEMA {{database_name}}.synapse_raw;
+-- Suspend ROOT TASK
+ALTER TASK IF EXISTS REFRESH_SYNAPSE_WAREHOUSE_S3_STAGE_TASK SUSPEND;
+-- Drop LATEST_TABLE UPSERTING TASK
+DROP TASK IF EXISTS UPSERT_TO_USERPROFILE_LATEST_TASK;
```
jaymedina, line 6: I realized that adding `IF EXISTS` is probably not the best idea because schemachange will silently not drop anything if there is a typo in the task name for example. The same goes for any other commands with the `IF EXISTS` conditional

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
```diff
@@ -0,0 +1,49 @@
...
+) 
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    COMMENT='This dynamic table contain the latest snapshot of user-profiles during the past 14 days. Snapshots are taken when user profiles are created or modified. Note: Snapshots are also taken periodically and independently of the changes. The snapshot_timestamp records when the snapshot was taken.'
+    AS 
+    WITH dedup_userprofile AS (
+        SELECT
+            *,
+            "row_number"()
```
philerooski, line 30: nit: you don't need to quote [this function](https://docs.snowflake.com/en/sql-reference/functions/row_number).

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
```diff
@@ -0,0 +1,49 @@
...
+                    PARTITION BY ID
+                    ORDER BY CHANGE_TIMESTAMP DESC, SNAPSHOT_TIMESTAMP DESC
+                )
+                AS N
+        FROM {{database_name}}.SYNAPSE_RAW.USERPROFILESNAPSHOT --noqa: TMP
+        WHERE
+            SNAPSHOT_DATE >= CURRENT_TIMESTAMP - INTERVAL '14 days'
+        QUALIFY
+            N=1
+    )
```
philerooski, line 41: This works fine, but there's an even cleaner way to do this that doesn't require you to add `row_number` as a column `N` and then later drop it. You can see an example of how Jenny does this for the `node_latest` dynamic table conversion [here](https://github.com/Sage-Bionetworks/snowflake/pull/89/files#diff-9e0dcdf9d925d9de6e0da98f08cdbf93df569a49133cb2a97932d507bcf43948R7-R18).

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
```diff
@@ -0,0 +1,49 @@
...
+    SELECT 
+        * exclude (N, LOCATION, COMPANY, POSITION, INDUSTRY), 
+        NULLIF(LOCATION, '') AS LOCATION, 
+        NULLIF(COMPANY, '') AS COMPANY, 
+        NULLIF(POSITION, '') AS POSITION, 
+        NULLIF(INDUSTRY, '') AS INDUSTRY,
```
philerooski, line 47: Did we ever figure out why some columns were being recorded as null and others as a blank string?

philerooski (approved):

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
philerooski: If you're not modifying any columns from the upstream table I recommend using the more succinct style which you can see in Jenny's PR which I linked to below.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
danlu1: Nope, not sure why this happened. Besides these columns, First_Name and Last_Name do have both empty strings and nulls. But I feel these 4 columns are more generic for business metrics so I convert empty strings for them.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
danlu1: Good point. I will add column comments as a separate step as like what Jenny said, new columns might be added to the source table.

synapse_data_warehouse/synapse_raw/tasks/V2.29.1__delete_userprofile_latest_tasks.sql
danlu1: Will drop IF EXISTS. Thanks for catching this.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
jaymedina: There's a platform ticket to investigate this [here](https://sagebionetworks.jira.com/browse/SWC-7215?atlOrigin=eyJpIjoiNGY5YjhjMWJmYzJiNGZkNDg2NThkYjhmMDljMzUzYzYiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

thomasyu888:
Depends on: https://github.com/Sage-Bionetworks/snowflake/pull/99 due to V script versioning

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
jaymedina: Just a heads up: I noticed that dynamic table comments don't show up in their own little Description box the way regular table comments do. With the dynamic table, you have to dive into the **Table Definition** itself to see that the dynamic table comment was added. It would be nice if we can also have the dynamic table comments front and center the way it is with the normal ones, but for now this is what it is. Dynamic tables were just added last year, so I'm sure this is a QOL improvement they'll add in the future.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE OR REPLACE DYNAMIC TABLE USERPROFILE_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS 
+    WITH dedup_userprofile AS (
+        SELECT
+            *
+        FROM {{database_name}}.SYNAPSE_RAW.USERPROFILESNAPSHOT --noqa: TMP
+        WHERE
+            SNAPSHOT_DATE >= CURRENT_TIMESTAMP - INTERVAL '14 days'
```
thomasyu888, line 12: Should we be consistent and use the same day interval for reduplication of all snapshots?

jaymedina: Consistency is a good idea. If we do this, we will have to go with `INTERVAL '30 days'` as that is the time window for `node_latest` and we can't make that one shorter.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
philerooski: I bugged Marco about this in the Slack thread to see if we can figure out whether this is in the Parquet or if Snowflake is introducing these inconsistencies during data load.

@danlu1 if you want to merge this before we've figured out the root of the issue can you leave an inline comment with the Jira ticket identifier which Jenny linked so that we have some provenance for these changes?

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
danlu1: Yes, the only ways to check the Table comments are either hover over the table name and the comment section which is pretty tiny or refer to the Table Definition. Hope the Description box gets added soon.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
danlu1: I second Jenny. Another reason is that the majority of the existing dynamic table is implementing the 14-days interval.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
danlu1: As Marco's [reply](https://sagebionetworks.slack.com/archives/C05PBT2FUHF/p1735002126450889), it seems there is no easy answer for the discrepancies since he noticed this in Athena as well. I'd just leave a comment and we can revisit after SWC-7215 is resolved.

thomasyu888:
> The above PR is still in draft mode. How about I merge this one first?

Ill defer to @jaymedina and you to self-organize.  The issue is the V script versions, if we merge this first, then it would have to be modified to 2.28 and Jenny's PR would have to be modified to 2.29.

I am ok with waiting a bit - Jenny's PR seems almost done - but since you'll be travelling for conference on M-Th, we can help merge if it's not complete.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
jaymedina: Good idea, thank you!

jaymedina:
@danlu1 I still have some testing to do as well as a review for #99 before it can get merged, so would you mind renaming your scripts to `V2.28.X`? Then we can merge this in ASAP. Sorry for the initial hold-up! Feel free to add me as validator for this update so I can handle that for you next week.

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
thomasyu888: @jaymedina Another risk that we assume is if snapshotting doesn't occur for 2 weeks in a row.

That said - I don't think that happens.  We will need to determine all the tables that we need to be concerned about delete events (even if they don't all track that change type).  So do all of these below need to use interval of 30?

* Teams
* Node
* File
* ACL
* teammember?
* project settings?

jaymedina: It looks like

* AR snapshots
* User group
* Userprofile

As well ( although User group and Userprofile don't capture deletions )

Another thought I just had is a lot of these tables are related to each other (file entities in `node_latest` should also be present in `file_latest`) sooo maybe for consistency as far as data availability, having all the same windows is better for the users. It would be nice to expedite having `change_type=DELETE` be available for all snapshots with a delete option on Synapse so we don't have to worry about the time window thing.

danlu1: @jaymedina @thomasyu888 @philerooski 
I did a brief check and the following snapshot tables have `change_type=DELETE`: 

- ACCESSREQUIREMENTSNAPSHOTS
- ACLSNAPSHOTS
- FILESNAPSHOTS
- NODESNAPSHOTS
- PROJECTSETTINGSNAPSHOTS

Do we agree to keep time window for dynamic tables derived from snopshots with change_type=DELETE to 30 days and other dynamic tables to 14 days?

danlu1:
@jaymedina The `V2.28` has been taken to refresh userprofilesnapshot. If we want this PR to be merged after #99, I can rename it as `V2.30.X`

synapse_data_warehouse/synapse/dynamic_tables/V2.29.4__userprofile_latest.sql
jaymedina: Yes! This workaround will do until platform team can address the missing `change_type=DELETE` for some snapshot tables. Thanks!

jaymedina:
@danlu1 - You're right - I merged my V2.29 scripts in #99. Feel free to rename your scripts to V2.30.X. I'm approving now so you can merge whenever that's done, and I've added myself as validator which I'll start on Monday.

Thank you for these updates!

jaymedina (approved):

PR #104, opened 2025-01-16
[SNOW-198] Create ownership database roles for data warehouse

philerooski:
### `synapse_data_warehouse/database_roles/V2.31.0__database_access_roles.sql`

* Since these are database-level objects, as opposed to account-level, we create these database roles and make grants on them within a versioned script.
* This is the first time we are creating a database-level object which is not a schema. Normally, a versioned script would exist within its associated schema folder, like `synapse_data_warehouse/synapse/...`. But since database roles are database-level objects, I write them to their own folder under `synapse_data_warehouse/database_roles`. I would like to move  `synapse_data_warehouse/synapse` and `synapse_data_warehouse/synapse_raw` to `synapse_data_warehouse/schemas/...` for uniformity, but I'll need to check if this works well with the `synapse_data_warehouse/schemachange-config.yml` file. Let me know if you prefer the `database_roles` folder alongside of the schema folders like it currently is in the PR.

### `admin/grants.sql`
* Although SYSADMIN has OWNERSHIP privileges on most or all of the objects having their ownership transferred here, we need MANAGE GRANTS privileges to COPY CURRENT GRANTS along with ownership.

### `.github/workflows/ci.yaml` and `admin/future_grants/V1.2.0__data_warehouse_ownership.sql`
* We want to make future grants as part of a versioned script, rather than revoking and reapplying them each time as part of `admin/grants.sql`.


synapse_data_warehouse/database_roles/V2.31.0__database_access_roles.sql
```diff
@@ -0,0 +1,22 @@
+USE DATABASE {{ database_name }}; --noqa: JJ01,PRS,TMP
+
+-- Create database roles which will own the respective namespace's objects
+CREATE OR REPLACE DATABASE ROLE SYNAPSE_ALL_ADMIN;
+CREATE OR REPLACE DATABASE ROLE SYNAPSE_RAW_ALL_ADMIN;
+CREATE OR REPLACE DATABASE ROLE SCHEMACHANGE_ALL_ADMIN;
```
philerooski, line 6: I use `CREATE OR REPLACE` instead of `CREATE IF NOT EXISTS` because if we were to rollback/replay these changes, we would want our database roles to be privilege-free.

thomasyu888 (commented):
🔥 this is looking great!

admin/grants.sql
```diff
@@ -498,3 +498,188 @@ GRANT ROLE SYNAPSE_DATA_WAREHOUSE_ANALYST TO SYNAPSE_DATA_WAREHOUSE_ADMIN;
...
+	TO DATABASE ROLE SYNAPSE_DATA_WAREHOUSE.SYNAPSE_ALL_ADMIN
+	COPY CURRENT GRANTS;
+GRANT OWNERSHIP ON FUTURE VIEWS
+	IN SCHEMA SYNAPSE_DATA_WAREHOUSE.SYNAPSE
+	TO DATABASE ROLE SYNAPSE_DATA_WAREHOUSE.SYNAPSE_ALL_ADMIN;
+GRANT OWNERSHIP ON TASKS
+	IN SCHEMA SYNAPSE_DATA_WAREHOUSE.SYNAPSE
+	TO DATABASE ROLE SYNAPSE_DATA_WAREHOUSE.SYNAPSE_ALL_ADMIN
+	COPY CURRENT GRANTS;
+GRANT OWNERSHIP ON FUTURE TASKS
```
jaymedina, line 545: What about future streams in case we make one in the `synapse` schema for whatever reason?

synapse_data_warehouse/database_roles/V2.31.0__database_access_roles.sql
jaymedina: Just to clarify: `synapse_data_warehouse_admin` (and its `dev` counterpart) is an account role, so this will be created manually, right? All the admin account roles are created manually?

jaymedina (commented):
This is looking good, I just have two comments:

1. The main thing I noticed is these grants are written for objects that already exist under these schemas, but the FUTURE grants don't account for objects that don't yet exist (e.g. dynamic tables in `synapse_raw`, streams in `synapse`).

I lean towards preemptively adding them in this PR, unless we're sure there's no way these kinds of objects will exist in those schemas. It's too bad there's no documentation for a `GRANT OWNERSHIP ON ALL FUTURE OBJECTS` command from what I've seen.

2. The `grants.sql` script is executed for every push to main ([here](https://github.com/Sage-Bionetworks/snowflake/blob/5acdecf3001a4145e377de5297fa58b0dd061159/.github/workflows/ci.yaml#L112)) and it seems odd to run these `FUTURE` grant lines every time there's a push to productio. I imagine this type of thing would live in a version script, that way we can keep track of how permissions change as our warehouse evolves.

philerooski:
Note to self:

```
GRANT OWNERSHIP ON TABLES
	IN SCHEMA SYNAPSE_DATA_WAREHOUSE.SYNAPSE_RAW
	TO DATABASE ROLE SYNAPSE_DATA_WAREHOUSE.SYNAPSE_RAW_ALL_ADMIN
	COPY CURRENT GRANTS;
```
and related `GRANT OWNERSHIP ON {object_type}S ...` is not the proper syntax. It should be `GRANT OWNERSHIP ON ALL {object_type}S ...`

We cannot grant future grants on object types which already have future grants granted upon them. But we can revoke future grants even if those future grants have not been granted on that object type, or even if the future grants have been granted to one role and we try to revoke them from another role (`REVOKE ... ON FUTURE ...` fails silently in that case). So we can prefix the future grants with `REVOKE ... ON FUTURE ...` statements to ensure that the future grants execute successfully.

admin/grants.sql
```diff
@@ -498,3 +498,188 @@ GRANT ROLE SYNAPSE_DATA_WAREHOUSE_ANALYST TO SYNAPSE_DATA_WAREHOUSE_ADMIN;
...
+
+---- RBAC reconfiguration of data warehouse ----
+-- The following grants transfer ownership of current and future 
+-- data warehouse objects from SYSADMIN to each namespace's respective
+-- `*_ALL_ADMIN` database role.
+
+---- SYNAPSE_DATA_WAREHOUSE ----
+
+-- SYNAPSE
+GRANT OWNERSHIP ON SCHEMA SYNAPSE_DATA_WAREHOUSE.SYNAPSE
```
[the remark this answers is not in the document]
philerooski: Yes, this works. We can use `... COPY CURRENT GRANTS` to keep everything the same except for the OWNERSHIP privilege. Even the future grants copy over.

thomasyu888 (approved):
🔥 LGTM!

philerooski:
@jaymedina 
> The main thing I noticed is these grants are written for objects that already exist under these schemas, but the FUTURE grants don't account for objects that don't yet exist... I lean towards preemptively adding them in this PR, unless we're sure there's no way these kinds of objects will exist in those schemas.

If we could be certain that the current object types which Snowflake makes available to us are all that would ever exist, I would agree with you. But I don't want a situation where Snowflake introduces a new object type (as recently happened with DYNAMIC TABLE) and the onus falls on the developer who is introducing a new object type within a namespace to figure out whether that object type existed when this PR was merged a.k.a. had already had future grants made upon it. My preferred M.O. is for developers to handle both grants and future grants when introducing a new object type within a namespace.

> As you mentioned, grants.sql script is executed for every push to main ([here](https://github.com/Sage-Bionetworks/snowflake/blob/5acdecf3001a4145e377de5297fa58b0dd061159/.github/workflows/ci.yaml#L112)), and it seems odd to run these FUTURE grant lines every time there's a push to production. I imagine this type of thing would live in a version script, that way we can keep track of how permissions change as our warehouse evolves.

I agree, it is inelegant to repeatedly apply these future grants. Especially when we need to preface them with a `REVOKE ... ON FUTURE ...` statement to get them to be idempotent. Initially I thought we didn't have an existing way to run versioned scripts as a `SECURITYADMIN`, but now I see it's actually pretty easy and that we are already doing this for warehouses. I'll move these future grant commands to versioned scripts in a `admin/future_grants/` directory. I tested that we can still `GRANT OWNERSHIP ON ALL ...` to a different role than the one specified in the future grant, so there's no side effects if we only run the future grants once, rather than concomitant with the `GRANT OWNERSHIP ON ALL ...` statements in grants.sql.

philerooski:
Blocked by #105

admin/grants.sql
philerooski: See my comment below.

.github/workflows/ci.yaml
```diff
@@ -146,6 +146,17 @@ jobs:
...
             -d METADATA
 
+      - name: future grants 
+        shell: bash
+        run: |
+          schemachange \
+            -f admin/future_grants \
+            -a $SNOWFLAKE_ACCOUNT \
+            -u $SNOWFLAKE_USER \
+            -r SECURITYADMIN \
```
jaymedina, line 156: I just logged onto the DPE_SERVICE account and noticed it has SYSADMIN access but no SECURITYADMIN access, so I feel compelled to ask: Are we sure the `$SNOWFLAKE_ACCOUNT` we use for these schemachange operations has access to SECURITYADMIN role? I don't see this role used elsewhere so asking just in case.

jaymedina (approved):
The recent changes LGTM, thanks! Just one comment/question.

jaymedina:
> _My preferred M.O. is for developers to handle both grants and future grants when introducing a new object type within a namespace._

Agreed, that seems cleaner. In that case, it's worth adding this to the contributions doc (e.g. "If you're adding a new object type, update the admin scripts with the appropriate grants and future grants for said object within the relevant namespace").

.github/workflows/ci.yaml
philerooski: Good catch. Looks like we've never had to use SECURITYADMIN to manage grants before. Everything has been done as the owner of an object granting privileges upon that object. I'll open another PR.

philerooski:
> In that case, it's worth adding this to the contributions doc (e.g. "If you're adding a new object type, update the admin scripts with the appropriate grants and future grants for said object within the relevant namespace").

I added [this section](https://github.com/Sage-Bionetworks/snowflake/pull/104/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R38) and included links to examples. But having written it all out I'm now realizing it's quite complex and there's some gray areas around edge cases. I think if you were already familiar with the functional/access model you could mimic the examples, but we might need proper documentation in the future.

PR #105, opened 2025-01-16
Allow SECURITYADMIN to deploy schemachange

philerooski:
From point 2 in [this comment](https://github.com/Sage-Bionetworks/snowflake/pull/104#issuecomment-2596909058), we want to execute future grants within a versioned script.

We already have a database, `METADATA`, which we use for managing the change history of account-level objects. Since future grants are grants on privileges which don't yet exist, we need to create them with a role which has the `MANAGE GRANTS` privilege (i.e., `SECURITYADMIN`). `SECURITYADMIN` doesn't yet have privileges to the `METADATA.SCHEMACHANGE.CHANGE_HISTORY` table. That's what this PR accomplishes.

While the object we are granting future privileges on are database and schema-level objects, rather than account-level objects, I still think it makes sense to track changes to their future grants in `METADATA` since its the future grants themselves which behave like account-level objects, requiring the account-level privilege `MANAGE GRANTS` to create.


thomasyu888 (approved):
🔥 LGTM!

jaymedina (approved):
Thanks for this!

PR #108, opened 2025-01-24
fix syntax in admin/grants.sql

philerooski:


linglp (approved):
LGTM! Thank you! Here's documentation of this syntax: https://docs.snowflake.com/en/sql-reference/sql/grant-role

PR #109, opened 2025-01-24
Grant SECURITYADMIN to service account

philerooski:


jaymedina (approved):

philerooski:
@jaymedina @thomasyu888 Apparently it's not as simple as adding a `GRANT ROLE SECURITYADMIN ...` to `grants.sql`. Normally the owner of a role or a role with `MANAGE GRANTS` privileges (a.k.a., just SECURITYADMIN by default) can grant that role, but SECURITYADMIN is a system-defined role and has no owner.

image

You'll notice in the above screenshot that ACCOUNTADMIN has a grant option on SECURITYADMIN. But granting the service account ACCOUNTADMIN just so that it can grant SECURITYADMIN or USERADMIN as part of the CI process seems like overkill.

I'm going to make an executive decision to grant SECURITYADMIN to the service account via the web UI, just so we don't have to bend over backwards and give the service account ACCOUNTADMIN to do it as part of the CI.

btw, this will also fix the sections in `admin.sql` where we were calling `USE ROLE SECURITYADMIN;`, not realizing that the service user never had privileges to assume SECURITYADMIN to begin with.

jaymedina:
This is fine with me 👍 I'm wondering why `securityadmin` wasn't one of the system-defined roles DPE_SERVICE automatically has access to in the first place.

PR #110, opened 2025-01-25
[SNOW-172] Convert file_latest to dynamic table

danlu1:
Problem:

Currently, the file_latest table is a table that is updated via upserting task and remove deleted records task that are based on filesnapshots stream. We want to convert it to dynamic table so it gets refreshed automatically.

Solution:

- [x] Add a script to create a backup of the original file_latest table which need to be removed manually after validation
- [x]  Add a script to delete the upserting task and deletion task for the original file_latest table
- [x]  Add a script to delete stream for filesnapshots table that is referred in file_latest upserting and deletion task
- [x]  Add a script to drop the original file_latest table so we can create a dynamic table with the same name
- [x]  Create file_latest dynamic table and filter out records that are non-preview or status as available.
- [x]  Add a script to add table and column comments.
- [x]  Explicitly SELECT the columns when creating dynamic table to avoid errors if upstream table changes. See https://github.com/Sage-Bionetworks/snowflake/pull/119 for more info.


synapse_data_warehouse/synapse_raw/streams/V2.33.2__drop_ filesnapshots_stream.sql
thomasyu888, line 1: Extra space in the filename

synapse_data_warehouse/synapse/dynamic_tables/V2.33.4__file_latest_dynamic_table.sql
```diff
@@ -0,0 +1,24 @@
...
+                PARTITION BY ID
+                ORDER BY CHANGE_TIMESTAMP DESC, SNAPSHOT_TIMESTAMP DESC
+            ) = 1
+    )
+    SELECT 
+        * 
+    FROM 
+        dedup_filesnapshots
+    WHERE
+        STATUS = 'AVAILABLE' OR NOT IS_PREVIEW;
```
thomasyu888, line 24: One comment, we currently provide analytics on the distribution of file handle statuses.  I think these would be great "gold layer" filters, but since we don't provide the raw tables to people, I think providing some of this info in the silver layer would still be valuable for some.

Most file statistic queries are done with joins against the node_latest table which makes it more rare to get a STATUS='UNLINKED' or IS_PREVIEW unless you query this table alone.

I think consulting platform about whether filehandles can be deleted will be valuable here.

philerooski: Agreed. We don't want to be filtering data beyond deduplication at this point. What if someone wants a dashboard that tells them how much data has been garbage-collected (STATUS = "ARCHIVED") or how much storage we currently dedicate to file previews (IS_PREVIEW = TRUE)? We don't want to have to build a dashboard from the bronze-level table.

synapse_data_warehouse/synapse/dynamic_tables/V2.33.5__add_column_comments_to_file_latest.sql
```diff
@@ -0,0 +1,25 @@
...
+COMMENT ON	COLUMN FILE_LATEST.CONTENT_SIZE IS 'The size of the file referenced by the file handle.';
+COMMENT ON	COLUMN FILE_LATEST.BUCKET IS 'The bucket where the file is physically stored. Applicable for s3 and GCP, otherwise empty.';
+COMMENT ON	COLUMN FILE_LATEST.KEY IS 'The key name uniquely identifies the object (file) in the bucket.';
+COMMENT ON	COLUMN FILE_LATEST.PREVIEW_ID IS 'The identifier of the file handle that contains a preview of the file referenced by this file handle.';
+COMMENT ON	COLUMN FILE_LATEST.IS_PREVIEW IS 'If true, the file referenced by this file handle is a preview of another file.';
+COMMENT ON	COLUMN FILE_LATEST.STATUS IS 'The availability status of the file referenced by the file handle. AVAILABLE: accessible via Synapse; UNLINKED: not referenced by Synapse and therefore available for garbage collection; ARCHIVED: the file has been garbage collected.';
+COMMENT ON	COLUMN FILE_LATEST.SNAPSHOT_DATE IS 'The data is partitioned for fast and cost effective queries. The snapshot_timestamp field is converted into a date and stored in the snapshot_date field for partitioning. The date should be used as a condition (WHERE CLAUSE) in the queries.';
```
philerooski, line 25: What happened to the formatting here? Can we remove the extra spaces?

danlu1:
Thanks guys for the feedback. @philerooski I keep the interval as 30 days here even though we don't filter out `CHANGE_TYPE != DELETE`. As what we discussed and archived [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3794894849/Converting+tables+to+dynamic+tables#:~:text=Defining%20the%20time,PROJECTSETTINGSNAPSHOTS.), for tables having CHANGE_TYPE with 'DELETE', we would use 30 days instead of 14 days.

thomasyu888 (approved):
🔥 LGTM! tagging @philerooski for final review.

synapse_data_warehouse/synapse/dynamic_tables/V2.33.4__file_latest_dynamic_table.sql
```diff
@@ -0,0 +1,22 @@
+-- Introduce the dynamic table
+USE SCHEMA {{database_name}}.synapse; --noqa: JJ01,PRS,TMP
+CREATE OR REPLACE DYNAMIC TABLE FILE_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS 
+    WITH dedup_filesnapshots AS (
+        SELECT
+            *
```
philerooski, line 9: You will need to explicitly name the columns here, so that when we add columns to the upstream table (SYNAPSE_RAW.FILESNAPSHOTS) the dynamic table doesn't break.

@jaymedina just made a PR to fix this for a dynamic table here: https://github.com/Sage-Bionetworks/snowflake/pull/119

danlu1: Thanks @philerooski . Another question I have is that I think it's better to rename the scripts to a later version instead of using v2.33.XXX since this is another attempt to update the table even though changes have already be pushed to dev database before. Does it sound good to you?

philerooski: @danlu1 Yes, good thinking. The revert wouldn't have dropped those records from the CHANGE_HISTORY table so you'll need to use a higher version.

PR #113, opened 2025-01-27
[SNOW-220] Revert ownership of tasks to sysadmin

philerooski:
Blocked by #112 

See [ticket](https://sagebionetworks.jira.com/browse/SNOW-220) for description.

I noticed that we have been granting privileges on non-existent objects in SYNAPSE_DATA_WAREHOUSE_DEV.SYNAPSE. I removed these from `grants.sql` and revoked future grants for these object types, as well.


jaymedina (approved):
I'm assuming leaving the `COPY CURRENT GRANTS` won't throw things off, right?

philerooski:
> I'm assuming leaving the COPY CURRENT GRANTS won't throw things off, right?

COPY CURRENT GRANTS is great because we can move the ownership around without affecting anyone else's privileges on these objects.

PR #116, opened 2025-01-29
[SNOW-221] Move ownership grants to versioned scripts

philerooski:
Transferring ownership, even to the role which already owns the task, will result in an auto-suspension of the task. My implementation here is to move the ownership grants to a versioned script, similar to what we recently did with future grants, so that they are only run once. Changes will be tracked along with other account-type changes in the `METADATA` database.


thomasyu888 (approved):
🔥 LGTM! Thanks for being on top of this.

admin/ownership_grants/V1.3.1__synapse_data_warehouse_prod.sql
```diff
@@ -0,0 +1,93 @@
...
+	COPY CURRENT GRANTS;
+GRANT OWNERSHIP
+	ON ALL STAGES
+	IN SCHEMA SYNAPSE_DATA_WAREHOUSE.SYNAPSE_RAW
+	TO DATABASE ROLE SYNAPSE_DATA_WAREHOUSE.SYNAPSE_RAW_ALL_ADMIN
+	COPY CURRENT GRANTS;
+GRANT OWNERSHIP
+	ON ALL STREAMS
+	IN SCHEMA SYNAPSE_DATA_WAREHOUSE.SYNAPSE_RAW
+	TO DATABASE ROLE SYNAPSE_DATA_WAREHOUSE.SYNAPSE_RAW_ALL_ADMIN
```
jaymedina, line 71: Would merging this be premature since we were discussing potentially granting ownership of everything to a separate database role?

philerooski: On streams, specifically? I don't think so. Streams are created on tables which exist in just one schema, so streams should be granted to the appropriate `*ALL_ADMIN` role.

The only objects which I foresee being affected by the multi-schema privilege work are the objects which we had to hand back to SYSADMIN recently (dynamic tables and tasks).

jaymedina: Got it! Thanks! Let me take another look through this then

jaymedina (approved):

PR #126, opened 2025-02-05
[SNOW-198] Update version of future grants script

philerooski:
From the discussion [here](https://sagebionetworks.jira.com/browse/SNOW-198?focusedCommentId=236657), the future grants were never applied as part of #104 since the version was not high enough.

I want to merge this directly into main to avoid the cherry-picking we had to do [last time](https://sagebionetworks.slack.com/archives/C02M7QR8D46/p1738684023276539). I will rebase dev on main after merge.


thomasyu888 (approved):
🔥 LGTM!

PR #131, opened 2025-02-06
[SNOW-218] Create proxy admin database role

philerooski:
### `synapse_data_warehouse/database_roles/V2.38.0__proxy_admin_role.sql` 
Create the proxy admin database role. Grant ownership and usage to the admin account role. (Will run in both dev and prod).

### `admin/ownership_grants/V1.6.0__proxy_admin_role_dev.sql` and `admin/ownership_grants/V1.6.1__proxy_admin_role_prod.sql`

Grant the proxy admin database role ownership and usage of `*ALL_ADMIN` database roles. Dev and prod, respectively.

### `admin/ownership_grants/V1.6.2__internamespace_object_ownership_dev.sql` and `admin/ownership_grants/V1.6.3__internamespace_object_ownership_prod.sql`

Transfer ownership of our internamespace objects from SYSADMIN to the proxy admin database role. Dev and prod, respectively.

### `admin/future_grants/V1.6.4__internamespace_object_ownership_dev.sql` and `admin/future_grants/V1.6.5__internamespace_object_ownership_prod.sql`

Grant ownership of future objects which potentially require internamespace privileges to the proxy admin database role. Dev and prod, respectively.

Depends on #133

PR #133, opened 2025-02-07
Remove/revoke erroneous grants

philerooski:
The changes here are primarily to address validation:

### `admin/grants.sql`
Remove some grants on object types which don't actually exist.

### `admin/future_grants/V1.6.0__remove_erroneous_grant.sql`
* Revoke future grants on tasks and dynamic tables. This is more of a stopgap until #131 is merged, at which point these object types will be owned by the proxy admin role. If any tasks or dynamic tables are deployed in the meanwhile, they will be owned by SYSADMIN, which is what we want (pending #131).
* Some future grants were made on object types which don't yet exist. We shouldn't yet be making these future grants. Some roles will be removed in #134. For example, we have a database role `SYNAPSE_DATA_WAREHOUSE_DEV.SYNAPSE_STAGE_READ` despite not having any stages in `SYNAPSE_DATA_WAREHOUSE_DEV.SYNAPSE`.


thomasyu888 (approved):
🔥 LGTM. Will wait for jenny to review before I merge

jaymedina (approved):
LGTM!

PR #140, opened 2025-02-13
[SNOW-147] Filter out deleted ACLs from `ACL_LATEST`

jaymedina:
## problem

`aclsnapshots` does not capture deleted ACLs, and there was an issue in the way the latest ACLs were retrieved. Both are addressed in the solution below.

## solution

- [x] Remove R script creating the `ACL_LATEST` dynamic table
- [x] Add V script re-introducing the `ACL_LATEST` dynamic table, refactored
- [x] Add a `snapshot_date` window to filter out for only the latest ACLs, which should minimize the risk of including deleted ACLs

### Original Process
1. **Unpack**
2. **Dedup**

### Original Data Structure

In this example, principal ID 200 gets `READ` privileges removed in 2025.
| owner-id | record-access                                       | timestamp    |
|----------|-----------------------------------------------------|--------------|
| 2        | {100: [read, download], 200: [read]}                | 12-1-2024     |
| 2        | {100: [read, download]}                             | 1-1-2025   |

### Data Structure (After Unpacking & Deduplicating)

Notice how unpacking before deduplicating, and defining a duplicate as a repeated entry of `owner_id` `access_type` and `prinicpal_id`, lets the ACL with pid 200 sneak into the final latest table, even though their permissions were removed in 1-1-2025.

| owner-id | access-type | pid | timestamp    |
|----------|-------------|-----|--------------|
| 2        | read        | 100 | 1-1-2025     |
| 2        | download    | 100 | 1-1-2025     |
| 2        | read        | 200 | 12-1-2024    |

---

### New Process
1. **Dedup**
2. **Unpack**

### Data Structure (After Deduplicating & Unpacking)

When we first deduplicate only based on the `owner_id`, and _then_ unpack the `access_type` and `principal_id`, only the ACL from 1-1-2025 is kept in the latest table.

| owner-id | access-type | pid | timestamp    |
|----------|-------------|-----|--------------|
| 2        | read        | 100 | 1-1-2025     |
| 2        | download    | 100 | 1-1-2025     |

## testing

<details>
  <summary>For testing, I created a temporary table for `ACL_LATEST` that I ran all the queries against. </summary>

```
CREATE OR REPLACE TEMPORARY TABLE temp_acl_latest AS
WITH dedup_acl AS (
    SELECT
        *,
        parse_json(resource_access) AS acl
    FROM synapse_data_warehouse.SYNAPSE_RAW.ACLSNAPSHOTS
    WHERE
        SNAPSHOT_DATE >= CURRENT_TIMESTAMP - INTERVAL '14 days'
    QUALIFY
        ROW_NUMBER() OVER (
            PARTITION BY OWNER_ID
            ORDER BY CHANGE_TIMESTAMP DESC, SNAPSHOT_TIMESTAMP DESC
        ) = 1
),
dedup_acl_expanded AS (
    SELECT
        CREATED_ON,
        CHANGE_TIMESTAMP,
        CHANGE_TYPE,
        OWNER_ID,
        OWNER_TYPE,
        SNAPSHOT_DATE,
        SNAPSHOT_TIMESTAMP,
        COALESCE(
            array_sort(value:"accesstype"::variant),
            array_sort(value:"accessType"::variant),
            array_sort(value:"accesstype#1"::variant),
            array_sort(value:"accesstype#2"::variant),
            array_sort(value:"accesstype#3"::variant)
        ) AS access_type,
        COALESCE(
            value:"principalId"::number,
            value:"principalid"::number,
            value:"principalid#1"::number
        ) AS principal_id
    FROM 
        dedup_acl,
        LATERAL FLATTEN(acl, outer => TRUE)
)
SELECT
    *
FROM
    dedup_acl_expanded;
   ```
</details>

<details><summary>Next, I queried for the discrepancies between the temporary table and the current `acl_latest`</summary>

```
WITH temp_counts AS (
    SELECT owner_id, COUNT(accesstype) AS temp_count
    FROM temp_acl_latest
    GROUP BY owner_id
),
acl_counts AS (
    SELECT owner_id, COUNT(access_type) AS acl_count
    FROM synapse_data_warehouse.synapse.acl_latest
    GROUP BY owner_id
)
SELECT 
    COALESCE(t.owner_id, a.owner_id) AS owner_id,
    t.temp_count,
    a.acl_count,
    ABS(COALESCE(t.temp_count, 0) - COALESCE(a.acl_count, 0)) AS difference
FROM temp_counts t
FULL OUTER JOIN acl_counts a
    ON t.owner_id = a.owner_id
WHERE COALESCE(t.temp_count, 0) != COALESCE(a.acl_count, 0);
```

</details> 

<details><summary>Lastly, I would arbitrarily select entities (`owner_id`) from the previous query and compare the ACL results between the new solution table and the current ACL_LATEST table, against the entity itself on Synapse.</summary>

```
select owner_id, change_timestamp, access_type, principal_id
from temp_acl_latest 
where owner_id = 24246541
order by change_timestamp desc, principal_id desc;

select *
from synapse_data_warehouse.synapse.acl_latest
where owner_id = 24246541
order by change_timestamp desc, principal_id desc;
```
</details> 

<details><summary>Here are some example results that confirm the results from `temp_acl_latest` are more accurate</summary>

1. syn3482905 no longer exists, and ACL does not show up in new solution, but does in `acl_latest`

image
image

2. same with syn27899252

image
image

3. syn63997960 has 6 ACLs on synapse, matching the new solution

image

new sol:
image

original sol has 3 ACLs that were removed:
image

</details> 

<details><summary>Leveraging NODE_LATEST/TEAM_LATEST for further verification</summary>

I looked to see what nodes might be missing from the new solution, and found that all the ones listed are nodes that don't exist anymore (a separate problem for `node_latest`).

```
SELECT *
FROM synapse.node_latest
WHERE benefactor_id = id
  AND id NOT IN (
    SELECT DISTINCT owner_id
    FROM temp_acl_latest
  );
```
image

Likeswise, I did the same for teams and saw that MOST of the missing IDs are because they don't exist in the original `aclsnapshots` table OR because of the 14 day cutoff in our new solution (see below for the ones that ARE in aclsnapshots, but don't make it through the 14 day cutoff):

```
with ids_in_team_latest_not_in_acl_latest as (
select id
from synapse.team_latest
where id not in (
    SELECT DISTINCT owner_id
        FROM temp_acl_latest
      )
order by snapshot_timestamp desc, change_timestamp desc
)
select *
from synapse_raw.aclsnapshots
where owner_id in (select id from ids_in_team_latest_not_in_acl_latest);
```

image

</details>

### CODE REVIEW

**Comment by philerooski** [comment]
No differences between that interactive rebase I did and this new branch, so you're good on that front 👍

File: synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+                value:"principalid#1"::number
+            ) AS principal_id
+        FROM 
+            dedup_acl,
+            LATERAL FLATTEN(acl, outer=>TRUE)
+    )
+    SELECT
+        *
+    FROM
+        dedup_acl_expanded;
```
BryanFauble, line 49: I think you could drop this second CTE since right after you do it you're doing a `SELECT *` on it.

```suggestion
)
SELECT
    CHANGE_TIMESTAMP,
    CHANGE_TYPE,
    CREATED_ON,
    OWNER_ID,
    OWNER_TYPE,
    SNAPSHOT_DATE,
    SNAPSHOT_TIMESTAMP,
    COALESCE(
        array_sort(value:"accesstype"::variant),
        array_sort(value:"accessType"::variant),
        array_sort(value:"accesstype#1"::variant),
        array_sort(value:"accesstype#2"::variant),
        array_sort(value:"accesstype#3"::variant)
    ) AS access_type,
    COALESCE(
        value:"principalId"::number,
        value:"principalid"::number,
        value:"principalid#1"::number
    ) AS principal_id
FROM 
    dedup_acl,
    LATERAL FLATTEN(acl, outer=>TRUE)
```

BryanFauble (approved):
LGTM!

jaymedina:
Thanks! I'll be merging this once https://github.com/Sage-Bionetworks/snowflake/pull/142 merges into `main`.

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
```diff
@@ -0,0 +1,43 @@
...
+
+
+CREATE OR REPLACE DYNAMIC TABLE ACL_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS 
+    WITH dedup_acl AS (
+        SELECT
+            *,
+            parse_json(resource_access) as acl,
```
philerooski, line 11: Interpreting the fields in `resource_access` isn't necessary to determine the latest snapshot, so we're making Snowflake call `parse_json` unnecessarily over a pretty significant number of snapshots which we're just going to drop anyways.

Can we move `parse_json` into the `LATERAL FLATTEN` function instead?

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
```diff
@@ -0,0 +1,43 @@
...
+        array_sort(value:"accesstype"::variant),
+        array_sort(value:"accessType"::variant),
+        array_sort(value:"accesstype#1"::variant),
+        array_sort(value:"accesstype#2"::variant),
+        array_sort(value:"accesstype#3"::variant)
+    ) AS access_type,
+    COALESCE(
+        value:"principalId"::number,
+        value:"principalid"::number,
+        value:"principalid#1"::number
```
philerooski, line 39: Are these all possible field names in the `resource_access` JSON?

A normal looking `resource_access` looks like
```
[
  {
    "accessType": [
      "CREATE",
      "CHANGE_SETTINGS",
      "DELETE",
      "DOWNLOAD",
      "CHANGE_PERMISSIONS",
      "READ",
      "MODERATE",
      "UPDATE"
    ],
    "principalId": 3336429
  },
  ...
]
```
But we sometimes have fields like `accesstype#3` and `pricipalid#1` as well? How do we know we won't see a `accesstype#4` in the future?

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
```diff
@@ -0,0 +1,45 @@
...
+    AS 
+    WITH dedup_acl AS (
+        SELECT
+            *,
+            parse_json(resource_access) as acl,
+        FROM {{database_name}}.SYNAPSE_RAW.ACLSNAPSHOTS --noqa: TMP
+        WHERE
+            SNAPSHOT_DATE >= CURRENT_TIMESTAMP - INTERVAL '14 days'
+        AND
+            RESOURCE_ACCESS != '[]'
```
jaymedina, line 16: Based on the Slack thread [here](https://sagebionetworks.slack.com/archives/C05PBT2FUHF/p1739394419004849). I can confirm that are no more `NULL` values in the `acl_latest` table with this filter:

image

Before there were:

image
.

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
```diff
@@ -0,0 +1,51 @@
...
+            array_sort(value:"accesstype#3"::variant)
+        ) AS access_type,
+        COALESCE(
+            value:"principalId"::number,
+            value:"principalid"::number,
+            value:"principalid#1"::number
+        ) AS principal_id
+    FROM 
+        dedup_acl,
+        LATERAL FLATTEN(PARSE_JSON(RESOURCE_ACCESS));
```
jaymedina, line 51: We no longer need `outer=>true` because we are filtering out objects that cannot be exploded.

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
jaymedina: Resolved!

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
jaymedina: Great point - I'm working on a more elegant solution to this that does not involve hard-coding the field names

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
```diff
@@ -0,0 +1,85 @@
...
+            )) AS access_type, -- Grab the access type VARIANT programmatically regardless of field name
+            MAX(
+                CASE 
+                    WHEN key ILIKE '%principal%' THEN value::number
+                END
+            ) AS principal_id -- Grab the principal ID NUMBER programmatically regardless of field name
+        FROM 
+            dedup_acl_level1_unpack,
+            LATERAL FLATTEN(INPUT => ACL_ENTRY)
+        GROUP BY
```
thomasyu888, line 74: Nit: check the different `GROUP BY` functions that snowflake offers.

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
philerooski: This is a clever solution. Is this ('%access%') as precise as we can be to avoid matching some new field that might be introduced but that is semantically different from accessType? Can we use additional cases in the CASE...END statement to strike that balance between specificity and generalizability?

But before we put too much thought into that, we should acknowledge that this is an upstream bug/feature handed down to us from the Platform team. Fields like `accessType1` and `accessType2` are an obvious indicator that `accessType` should have been modeled as a list of objects. You may want to file a ticket with Platform asking them to put this data into a sane format. In the meanwhile, I'll leave it to you whether we should revert to the naive solution of selecting `accessType1`, `accessType2`, etc. explicitly. or whether we should maintain the pattern-matching solution. They each have their drawbacks.

jaymedina: > we should acknowledge that this is an upstream bug/feature handed down to us from the Platform team. Fields like accessType1 and accessType2 are an obvious indicator that accessType should have been modeled as a list of objects. You may want to file a ticket with Platform asking them to put this data into a sane format.

That's true, I'll file a ticket for them to address this before taking the snapshots. I anticipate that platform won't get to this any time soon while they are backed up with other items, so in the meantime, I'd like to move forward with this since it's a bit more robust than the previous solution.

I definitely think there are ways we can make this more airtight, but since the EL ACL reports are due by the end of this week, I'd like to move forward with the solution we have now and I can make a ticket to revisit later with the considerations below:

> Is this ('%access%') as precise as we can be to avoid matching some new field that might be introduced but that is semantically different from accessType? Can we use additional cases in the CASE...END statement to strike that balance between specificity and generalizability?

Let me know if you agree with this plan!

philerooski: Sounds good.

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
jaymedina: I looked at `GROUP BY ALL` and `GROUP BY GROUPING SETS (...)`:

`GROUP BY ALL` overgroups the results (we are expecting 614K rows)

image

and `GROUP BY GROUPING SETS (...)` is used when grouping by multiple categories, but here we are just leveraging `GROUP BY` to group by the `ACL_ENTRY` just to pair the access types with the right principal IDs. 

Is there another function you think I should look at, and was this for optimization or to simplyify the query?

jaymedina:
I just bumped up the latest version, would y'all mind giving this one last review before I merge? Thanks!

synapse_data_warehouse/synapse/dynamic_tables/V2.40.0__acl_latest_refactored.sql
jaymedina: [Ticket](https://sagebionetworks.jira.com/browse/SNOW-243) for revisiting this query for further optimization.

I'll make the platform ticket after lunch and resolve this thread when made.

synapse_data_warehouse/synapse/dynamic_tables/V2.41.0__acl_latest_refactored.sql
```diff
@@ -0,0 +1,85 @@
...
+            OWNER_ID,
+            OWNER_TYPE,
+            SNAPSHOT_DATE,
+            SNAPSHOT_TIMESTAMP,
+            RESOURCE_ACCESS
+        FROM {{ database_name }}.SYNAPSE_RAW.ACLSNAPSHOTS --noqa: TMP
+        WHERE
+            SNAPSHOT_DATE >= CURRENT_TIMESTAMP - INTERVAL '14 days'
+        AND
+            RESOURCE_ACCESS != '[]'
```
philerooski, line 25: If this isn't already documented in the ticket, can we get an inline comment explaining what it means for RESOURCE_ACCESS == '[]'

jaymedina: Added! https://github.com/Sage-Bionetworks/snowflake/pull/140/commits/ee794901799db2d01ca67a24ae3ecf03e2eb42e3

PR #143, opened 2025-02-14
[SNOW-89] Create admin and developer service users

philerooski:
I will alter the user in snowsight to configure the [key-pair authentication](https://docs.snowflake.com/en/user-guide/key-pair-auth#configuring-key-pair-authentication). Although it's possible for me to include the public key in the CREATE USER statement if we feel like that's worth encoding here.


admin/users.sql
```diff
@@ -124,3 +124,9 @@ CREATE USER IF NOT EXISTS AD_SERVICE
 CREATE USER IF NOT EXISTS RECOVER_SERVICE
     PASSWORD = 'placeholder'
     MUST_CHANGE_PASSWORD = TRUE;
+
+CREATE USER IF NOT EXISTS ADMIN_SERVICE
+    TYPE = SERVICE;
+
+CREATE USER IF NOT EXISTS DEVELOPER_SERVICE
```
jaymedina, line 131: Are we no longer going to use the `DPE_SERVICE` service account?

philerooski: See my comment [here](https://sagebionetworks.jira.com/browse/SNOW-89?focusedCommentId=237696). I thought that `DEVELOPER_SERVICE` was a more specific title for what this user is intended to do.

jaymedina (approved):

PR #144, opened 2025-02-14
[SNOW-89] Grant roles to service accounts

philerooski:
I hadn't realized I could have done this as part of #143


danlu1 (approved):
LGTM.

PR #146, opened 2025-02-17
[SNOW-89] Update authentication policy for service accounts

philerooski:
_How_ a user is allowed to authenticate is regulated by authentication policies. This PR alters our existing authentication policy for service accounts by additionally allowing key-pair authentication. I also correct a typo in the `integrations.sql` file.


admin/policies.sql
```diff
@@ -77,6 +77,10 @@ ALTER USER RECOVER_SERVICE SET AUTHENTICATION POLICY service_account_authenticat
...
 ALTER USER AD_SERVICE SET AUTHENTICATION POLICY service_account_authentication_policy;
 ALTER USER THOMASYU888 SET AUTHENTICATION POLICY service_account_authentication_policy;
+ALTER USER ADMIN_SERVICE SET AUTHENTICATION POLICY service_account_authentication_policy;
+ALTER USER DEVELOPER_SERVICE SET AUTHENTICATION POLICY service_account_authentication_policy;
 
-USE ROLE ACCOUNTADMIN;
 ALTER ACCOUNT SET CORTEX_ENABLED_CROSS_REGION = 'ANY_REGION';
+
+ALTER AUTHENTICATION POLICY service_account_authentication_policy
+  AUTHENTICATION_METHODS = ('PASSWORD', 'KEYPAIR');
```
thomasyu888, line 86: This is surprising. I thought the whole premise around service type accounts was that it only allowed keypair auth...

philerooski: @thomasyu888 That's correct. If the user type is "SERVICE" then you can only authenticate via key-pair. That doesn't mean that user gets a free pass from the existing authentication policies to authenticate using key-pair.

thomasyu888 (approved):
🔥 LGTM!

PR #149, opened 2025-02-19
[SNOW-177]Convert teammember_latest to dynamic table

danlu1:
Problem:

Currently, the teammember_latest table is a table that is updated via upserting task that are based on teammembersnapshots stream. We want to convert it to dynamic table so it gets refreshed automatically.

Solution:

 

- [x] Add a script to create a backup of the original teammember_latest table which need to be removed manually after validation
- [x] Add a script to delete the upserting task for the original teammember_latest table
- [x] Add a script to delete stream for teammembersnapshots table that is referred in teammember_latest upserting task
- [x] Add a script to drop the original teammember_latest table so we can create a dynamic table with the same name
- [x] Create teammember_latest dynamic table  and add table and column comments at the same time according to the newest SOP.


synapse_data_warehouse/synapse/dynamic_tables/V2.42.4__teammember_latest_dynamic_table.sql
```diff
@@ -0,0 +1,41 @@
...
+    )
+    SELECT 
+        CHANGE_TYPE,
+	    CHANGE_TIMESTAMP,
+	    CHANGE_USER_ID, 
+	    SNAPSHOT_TIMESTAMP, 
+	    TEAM_ID, 
+	    MEMBER_ID, 
+	    IS_ADMIN, 
+	    SNAPSHOT_DATE
```
philerooski, line 39: nit: formatting (extra tab?)

philerooski (approved):
Wait for others review before merging, but this looks good to me.

It also just occurred to me that we should probably be doing these dynamic table conversions in two PRs:

1. Creation of the dynamic table
2. (Validation)
3. Deletion of old table and related objects

This allows us to compare the upsert behavior of the dynamic table to that of the original table. That's something we don't get with a static backup of the original table.

synapse_data_warehouse/synapse/dynamic_tables/V2.42.4__teammember_latest_dynamic_table.sql
```diff
@@ -0,0 +1,41 @@
...
+	    IS_ADMIN BOOLEAN COMMENT 'If true, then the member is manager of the team.',
+	    SNAPSHOT_DATE DATE COMMENT 'The data is partitioned for fast and cost effective queries. The snapshot_timestamp field is converted into a date and stored in the snapshot_date field for partitioning. The date should be used as a condition (WHERE CLAUSE) in the queries.'
+    )
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    COMMENT = 'This table contain snapshots of team-members during the past 14 days. Snapshots are captured when a team and/or its members are modified. Note: Snapshots are also taken periodically and independently of the changes. The snapshot_timestamp records when the snapshot was taken.'
+    AS 
+    WITH dedup_teammembers AS (
+        SELECT
+            *
```
jaymedina, line 21: Do we want to explicitly call the columns in the first CTE as opposed to the final SELECT call, so that `ROW_NUMBER` sorting and deduplication isn't performed on extra columns?

I'm also remembering that a Snowflake developer said using `SELECT *` on an initial query will pull in _all_ columns, even those newly-introduced in the `*snapshots` table that don't yet exist in the `*latest` table. This may cause the dynamic table to break, so I'm thinking it's a good rule of thumb to explicitly call the column names in an initial `SELECT`, and then the main query (or subsequent CTEs) can use `SELECT *` from the _initial_ CTE. 

LMK what you think and I can update the SOP accordingly!

synapse_data_warehouse/synapse/dynamic_tables/V2.42.4__teammember_latest_dynamic_table.sql
danlu1: Thanks for catching this. Weird that I don't see extra tab locally.

synapse_data_warehouse/synapse/dynamic_tables/V2.42.4__teammember_latest_dynamic_table.sql
danlu1: I think the SOP is overall pretty good and clear. But I like Phil's idea of adjusting it to allow to compare dynamic table to the original table without time constrains by deleting old objects in a later PR.

jaymedina: > adjusting it to allow to compare dynamic table to the original table without time constrains

That's a good point, this would be valuable for validators. We only have 1 more dynamic table to fix, though, so I'm not sure if revisiting the processes is worth it (we have `CERTIFIEDQUIZ_LATEST` left)

danlu1: I think it might be interesting if we test the new process for the first and last time 👀

jaymedina: That works for me! Maybe this part of the process will be a good reference for future SOPs. I can update our SOP and tag you once it's ready for review. Did you want to also test it on this PR since it's not merged yet?

danlu1: Awesome. I think no need to test on this one since it's reviewed already. We can test it on `CERTIFIEDQUIZ_LATEST`.

PR #153, opened 2025-02-21
[SNOW-218] Additional privileges needed for the execution of tasks

philerooski:
It's not sufficient for a a task owner to have:

* OWNERSHIP on a task
* OWNERSHIP on a stage
* EXECUTE MANAGED TASK on the account

For a data loading task to execute successfully. Additionally, we need:

* EXECUTE TASK on the account
  - So that we are allowed to execute tasks _generally_
* USAGE on the security integration 
  - So that we can use external stages


BryanFauble (approved):
LGTM!

PR #156, opened 2025-02-25
[SNOW-224] Remove CI race condition

philerooski:
Currently, our CI job executes all its jobs at once, although the order in which those jobs runs may be important. I added `needs` fields to each job which should wait for the successful completion of another job before running. I also removed some unused, commented-out lines and reordered the jobs in the `ci.yaml` file itself so that they are ordered according to the order in which they will execute (this is why it looks like the changes are more drastic than they really are).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,8 +1,5 @@
-# Very basic GitHub Action workflow, goes in ~/.github/workflows/deploy.yaml
-
 name: CI
 
-# Controls when the workflow will run
 on:
   # Triggers the workflow on any branch or tag commit
   push:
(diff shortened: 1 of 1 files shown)


danlu1 (approved):
Thanks so much for working on it. It looks good to me. My understanding is the we will implement changes for objects in the datawarehouse (dev and prod) before fine tuning/ updating account level setting to avoid privilege issue when modifying the datawarehouse.

philerooski:
@danlu1 That's one way of looking at it. I tried to mirror the order in which schemachange applies versioned vs. repeatable scripts:

1. non-admin schemachange scripts
2. admin schemachange scripts
3. admin snowsql scripts

(3) is analogous to schemachange repeatable scripts, which normally run after schemachange versioned scripts, so we mirror that behavior here in the ordering 2 → 3.

PR #125, opened 2025-02-04
[SNOW-90] Introduce CI/CD pipeline

jaymedina:
## problem

Rather than have it be a manual process, automating the testing of `schemachange` updates in our branch before merging it into `dev` will speed up the development life cycle. A design document was made to tackle this problem and is available [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3703472153/SNOW-90+CI+Pipeline+for+Schemachange+Testing+in+Sage+s+Snowflake+Development+Environment?focusedCommentId=3706716206).

## solution

[See here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3703472153/SNOW-90+CI+Pipeline+for+Schemachange+Testing+in+Sage+s+Snowflake+Development+Environment#Recommendation) for the design document illustrating the final solution.

- [x] Introduce a new branch label, (`skip_cloning`), to opt-out of the automated process of zero-copy cloning `synapse_data_warehouse_dev` and testing new/modified schemachange scripts against it
- [x] Introduce new CI jobs that reflect the design outlined in the [design document](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3703472153/SNOW-90+CI+Pipeline+for+Schemachange+Testing+in+Sage+s+Snowflake+Development+Environment?focusedCommentId=3706716206)
- [x] Update contribution guidelines to explain these new CI jobs and the corresponding PR labels

## testing

- [x] Test that label `skip_cloning` skips the CI/CD clone job (**branch**: this one, **status**: [passed](https://github.com/Sage-Bionetworks/snowflake/actions/runs/14134904400/job/39604163839)) 
- [x] Test that a clone is made of the `dev` db and _not_ the `prod` db (**branch**: this one, **status**: passed) 
- [x] Test that a clone is named after the branch (**branch**: this one, **status**: passed)
- [x] Test that merging a PR into _another_ PR triggers `create_clone_and_run_schemachange` (**branch**: ci-cd-test-merge, **status**: [passed](https://github.com/Sage-Bionetworks/snowflake/actions/runs/14136848823))

<details><summary>and changes show up in the clone for the PR that got merged into</summary>
image

</details> 

- [x] Test that merging a PR into _another_ PR with `skip_cloning` DOES NOT trigger `create_clone_and_run_schemachange` (**branch**: snow-90-update-docs, **status**: [passed](https://github.com/Sage-Bionetworks/snowflake/actions/runs/14137839731))
[...]

PR #157, opened 2025-02-27
[SNOW-89] Update CI with service users

philerooski:
This PR contains a significant refactor, although a small number of actual changes to how our CI works.

* `.github/actions/configure-snowflake-cli/action.yml`
  - Installing Snowflake clients and configuring their authentication is a common step in our CI jobs. This file consolidates these steps into a single file to reduce the redundancy in our primary CI workflow. More specifically, this action:
    + Writes a `config.toml` and `connections.toml` files to the `~/.snowflake/` directory.
    + Installs both the Snowflake CLI and schemachange
    + Verifies that our authentication is set up correctly
    + Takes ~20 seconds to run
* `.github/workflows/ci.yaml`
  - This is our primary CI workflow. Its behavior remains mostly the same as before, but has been significantly refactored. Behavior changes are related to the users and roles with which we deploy resources.
  - Deployment (by both schemachange and Snowflake CLI) to dev and prod is now done by the `ADMIN_SERVICE` user.
  - Schemachange deployment to dev is done using the SYNAPSE_DATA_WAREHOUSE_DEV_ADMIN role.
  - Schemachange deployment to prod is done using the SYNAPSE_DATA_WAREHOUSE_ADMIN role.
 * `admin/integrations.sql`
   - Snowflake recommends a new way of formatting command-line variables, which we've updated here.


philerooski:
I will squash before merging.

.github/workflows/ci.yaml
```diff
@@ -20,181 +20,147 @@ on:
...
-          python-version: '3.10'
+      - name: Checkout repository
+        uses: actions/checkout@v4
 
-      - name: install-py-dependencies
-        shell: bash
-        run: |
-          pip install schemachange==3.6.1
-          pip install numpy==1.26.4
-          pip install pandas==1.5.3
```
jaymedina, line 45: I think we'll need to add these back - these version locks were introduced because of failures after merging to `dev`: https://github.com/Sage-Bionetworks/snowflake/pull/65

philerooski: The CI at that point in time used `schemachange==3.6.1`, whereas I upgraded us to the most recent `schemachange==4.0.1`, so I'm not sure that these exact dependency versions still work (they are sub-dependencies so it's not obvious from looking at schemachange's requirements.txt).

Do you know why pinning the versions was necessary in the first place? The ticket itself was sparse on details, but I think if schemachange has their requirements.txt in good form then we don't need to worry about pinning the dependencies.

jaymedina:
> Installing Snowflake clients and configuring their authentication is a common step in our CI jobs. This file consolidates these steps into a single file to reduce the redundancy in our primary CI workflow.

This is great, and I'm assuming you'd want this merged into `dev` before https://github.com/Sage-Bionetworks/snowflake/pull/125 is that correct? If so, I would consider this a blocker to SNOW-90 and can shift my focus to finishing this review. Can we set a goal together to have this merged by tonight and do you think that's reasonable?

.github/actions/configure-snowflake-cli/action.yml
```diff
@@ -0,0 +1,110 @@
...
+        mkdir -p $HOME/.snowflake
+        CONFIG_FILE=$HOME/.snowflake/config.toml
+        CONNECTION_FILE=$HOME/.snowflake/connections.toml
+        PRIVATE_KEY_FILE=$HOME/.snowflake/rsa_key.p8
+      
+        # Write to connections.toml file
+        echo '[default]' >> $CONNECTION_FILE
+        echo "account = \"${{ inputs.ACCOUNT }}\"" >> $CONNECTION_FILE
+        echo "user = \"${{ inputs.USER }}\"" >> $CONNECTION_FILE
+        echo 'warehouse = "${{ inputs.WAREHOUSE }}"' >> $CONNECTION_FILE
```
jaymedina, line 53: Syntax issue? `\"${{ inputs.WAREHOUSE }}\"`

.github/actions/configure-snowflake-cli/action.yml
```diff
@@ -0,0 +1,110 @@
...
+  connection:
+    description: >
+      A pretty-printed description of the default Snowflake connection
+      encoded using base64. Decode using `base64 --decode`.
+    value: ${{ steps.connection-test.outputs.connection }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Create config.toml file
```
jaymedina, line 39: nit: "Create config.toml and connections.toml files"

.github/actions/configure-snowflake-cli/action.yml
```diff
@@ -0,0 +1,110 @@
...
+        # Write to the private key file
+        printf "%s" "${{ inputs.PRIVATE_KEY }}" > $PRIVATE_KEY_FILE
+            
+    - name: Verify configuration/connection file contents
+      shell: bash
+      run: |
+        echo "Snowflake configuration is located at $CONFIG_FILE"
+        cat $CONFIG_FILE
+        echo "Snowflake connection is located at $CONNECTION_FILE"
+        cat $CONNECTION_FILE
```
jaymedina, line 85: Is it a security risk to be exposing the contents of the connection files this way? We have `PRIVATE_KEY_PASSPHRASE` in there

.github/workflows/ci.yaml
```diff
@@ -20,181 +20,147 @@ on:
...
+      - name: Configure Snowflake
+        uses: ./.github/actions/configure-snowflake-cli
+        with:
+          PRIVATE_KEY_PASSPHRASE: ${{ secrets.ADMIN_SERVICE_PASS_PHRASE }}
+          PRIVATE_KEY: ${{ secrets.ADMIN_SERVICE_PRIVATE_KEY }}
+          ACCOUNT: ${{ secrets.SNOWSQL_ACCOUNT }}
```
jaymedina, line 34: To be sure: Is this `SNOWSQL_ACCOUNT` the new `ADMIN_SERVICE` account or Tom's account still?

.github/workflows/ci.yaml
```diff
@@ -20,181 +20,147 @@ on:
...
+          STACK: ${{ vars.STACK }}
         run: |
           schemachange \
-            -f synapse_data_warehouse \
-            -a $SNOWFLAKE_ACCOUNT \
-            -u $SNOWFLAKE_USER \
-            -r SYSADMIN \
-            -w compute_xsmall \
-            --config-folder synapse_data_warehouse
+            --connection-name default \
```
jaymedina, line 47: This complicates the CI/CD deployment step in #125 a little, and I'd like to think out loud and have a conversation about it before moving forward to see if we can simplify this (but if not, that's okay).

In our current CI/CD workflow there are two connections involved:

1. connection to admin_service (to create and manage the clone)
2. connection to dpe_service (/developer_service) (to run schemachange)

Currently the way I have things set up is by configuring those 2 connections in a single config file and swapping back and forth between them depending on the step (steps 1 & 2 are done in the same job).

The changes introduced here will impact the CI/CD workflow in the following ways:

1. Each step will need its own job with its own config file generated based on the inputs
2. _Because_ admin_service and developer_service will have their own jobs, but developer_service relies on environment variables created in the admin_service job (namely, the new sanitized clone name to deploy changes to), I will need to rely on artifacts or outputs to "transfer" that new environment variable from the admin_service job to developer_service job. 

----

All in all, I can work with these changes as they currently are. But my question is, can we simplify #125 by adding support for multiple connections within the same config file in this PR?

jaymedina (commented):
This is excellent, thanks for the refactor! These are all my comments from an initial pass through. Most importantly I had comments on how this refactor will impact #125.

philerooski:
@jaymedina 

> I'm assuming you'd want this merged into dev before https://github.com/Sage-Bionetworks/snowflake/pull/125 is that correct? 

Not necessarily. I opened this PR into main, actually, since normally workflow changes only take effect after being merged into main. But after fact-checking myself I'm finding out that I probably only need to merge into the _default_ branch, which currently is configured to be the dev branch. I may need to rebase.

> If so, I would consider this a blocker to SNOW-90

Whichever is merged first, there will need to be a merge conflict resolution done for whichever is merged second. But the existing CI process will still work: the work you're doing for SNOW-90 doesn't need to use the new service accounts or Snowflake configuration action right away.

.github/actions/configure-snowflake-cli/action.yml
```diff
@@ -0,0 +1,110 @@
...
+  ACCOUNT:
+    description: "Snowflake account identifier"
+    required: true
+  USER:
+    description: "Snowflake user identifier"
+    required: true
+  WAREHOUSE:
+    description: "Snowflake warehouse"
+    required: false
+    default: ""
```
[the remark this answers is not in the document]
philerooski: That makes sense. I'll update it.

.github/actions/configure-snowflake-cli/action.yml
philerooski: I'm not sure if we need to use double quotes throughout the whole expression to properly evaluate `${{ inputs.WAREHOUSE }}`, but it wouldn't hurt.

.github/actions/configure-snowflake-cli/action.yml
philerooski: It's a secret, so Github censors the output. You can see how this is done here: https://github.com/Sage-Bionetworks/snowflake/actions/runs/13575165696/job/37949533375#step:3:85

.github/workflows/ci.yaml
philerooski: An _account_ in Snowflake means the Snowflake account. ADMIN_SERVICE and Tom's "account" are actually users. So this secret contains our Snowflake account identifier. I'm not sure why Tom labeled `SNOWSQL_ACCOUNT` with the client he was using to connect to the account.

.github/workflows/ci.yaml
```diff
@@ -20,181 +20,147 @@ on:
...
-          pip install numpy==1.26.4
-          pip install pandas==1.5.3
+      - name: Configure Snowflake
+        uses: ./.github/actions/configure-snowflake-cli
+        with:
+          PRIVATE_KEY_PASSPHRASE: ${{ secrets.ADMIN_SERVICE_PASS_PHRASE }}
+          PRIVATE_KEY: ${{ secrets.ADMIN_SERVICE_PRIVATE_KEY }}
+          ACCOUNT: ${{ secrets.SNOWSQL_ACCOUNT }}
+          USER: ${{ vars.ADMIN_SERVICE_USER }}
+          WAREHOUSE: ${{ secrets.SNOWSQL_WAREHOUSE }}
```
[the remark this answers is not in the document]
philerooski: It doesn't, that's just the way Tom set it up.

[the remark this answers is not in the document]
philerooski: I should probably replace this with either nothing (remove it, now that we have a default warehouse) or specify the warehouse explicitly. That would get rid of the unnecessary mystique.

.github/workflows/ci.yaml
philerooski: Yes, I think it would be possible, but we would need to decide if the juice is worth the squeeze.

In terms of implementation, the most straightforward thing would be to parameterize the configure-snowflake-cli action like `PRIVATE_KEY`, `PRIVATE_KEY_2`, etc., then double up the echo statements.

Although if we go this route we can also consider the more general problem of supporting multiple users/connections (the original design had one user for ACCOUNTADMIN, one for {DATABASE}_ADMIN and another for DEVELOPER, for example), which would probably mean passing a JSON blob as a parameter. That feels a little messy with the formatting of passing secrets and such.

I'm in favor of simplicity here. It's annoying that we need to go through the installation of the clients once again and pass variables as outputs rather than setting and getting them directly from the GITHUB_ENV, but from a conceptual/maintenance stand point I like the idea of keeping the action a little simplistic and not trying to tailor it too much to our use case.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
philerooski: I removed the argument in d8cbffd

.github/workflows/ci.yaml
philerooski: @jaymedina You can see a successful run of schemachange within a test workflow that uses the new action [here](https://github.com/Sage-Bionetworks/snowflake/actions/runs/13660879625/job/38191489616#step:5:1). It mirrors how we run schemachange in this PR, except it points at an "empty" analytics folder and my clone database so that it doesn't accidentally deploy anything to dev/prod. 

Is there a specific change [since 3.6.1](https://github.com/Snowflake-Labs/schemachange/releases) which you think we should be worried about? Downgrading at this point would be a big pain since connections.toml wasn't supported until [v4.0.0](https://github.com/Snowflake-Labs/schemachange/releases/tag/v4.0.0). I would need to completely rewrite the way auth is done to work specifically with schemachange (v3.6.1 only supports passing connection parameters via the schemachange-config.yml file or command-line parameters). With support for connections.toml, we have a uniform way to configure authentication across all of our clients: schemachange, Snowflake CLI, and [even the python connector](https://github.com/Snowflake-Labs/schemachange/issues/281) if we ever want to use that.

jaymedina: > Downgrading at this point would be a big pain since connections.toml wasn't supported until [v4.0.0](https://github.com/Snowflake-Labs/schemachange/releases/tag/v4.0.0).

Ah, this explains the bump up. I still think we should eventually have a lock file to version lock some of these dependencies. Since that issue happened with numPy/Pandas, there's no guarantee it won't happen again unless we lock those versions. How about we lock the dependencies after this merges into `main` and we're sure the current versions of everything work with no issues?

jaymedina:
[Here](https://sagebionetworks.jira.com/browse/SNOW-254) is the new ticket for refactoring CI/CD with this new action.

.github/workflows/ci.yaml
philerooski: numpy and pandas are not dependencies of v4.0.1. Here's the pip freeze:

```
asn1crypto==1.5.1
certifi==2025.1.31
cffi==1.17.1
charset-normalizer==3.4.1
cryptography==44.0.2
filelock==3.[17](https://github.com/Sage-Bionetworks/snowflake/actions/runs/13684265553/job/38263809921#step:5:18).0
idna==3.10
Jinja2==3.1.5
MarkupSafe==3.0.2
packaging==24.2
platformdirs==4.3.6
pycparser==2.22
PyJWT==2.10.1
pyOpenSSL==25.0.0
pytz==[20](https://github.com/Sage-Bionetworks/snowflake/actions/runs/13684265553/job/38263809921#step:5:21)25.1
PyYAML==6.0.2
requests==2.32.3
schemachange==4.0.1
snowflake-connector-python==3.14.0
sortedcontainers==2.4.0
structlog==[24](https://github.com/Sage-Bionetworks/snowflake/actions/runs/13684265553/job/38263809921#step:5:25).1.0
tomlkit==0.13.2
typing_extensions==4.12.2
urllib3==2.3.0
```

jaymedina: Hmm that's strange. Anyway, numPy and Pandas aren't my only concern, my concern is running into dependency issues in the future. What are your thoughts on this?

> How about we lock the dependencies after this merges into main and we're sure the current versions of everything work with no issues?

.github/workflows/ci.yaml
jaymedina: Got it!

.github/workflows/ci.yaml
jaymedina: > I like the idea of keeping the action a little simplistic and not trying to tailor it too much to our use case.

I can agree with this. I can work around this by passing outputs from job to job. If there are any obstacles we'll cross that bridge when we get there.

jaymedina (approved):
LGTM! 🔒 

Just wanted to close the loop on one last thread, but this looks ready for merge.

.github/workflows/ci.yaml
philerooski: Yeah, I think this is because pandas was removed as a dependency in v3.7.0.

> my concern is running into dependency issues in the future. What are your thoughts on this?

From my understanding, dependencies fall into two general categories:

1. Package (Python) dependencies
2. System dependencies

The most common issue I've run into in the past is with system dependencies. What usually happens is that the VM's OS updates one of the system libraries, introducing a breaking change into a package dependency that wasn't there previously. Besides pinning the OS, e.g., using ubuntu-24.04 rather than ubuntu-latest, there's not much we can do to avoid these types of breaking changes.

A slightly less common issue is with broken package dependencies. This tends to happen when package maintainers incorrectly configure their requirements.txt so that it's possible to install dependency versions which aren't compatible with the package. A typical workaround for this would be to manually override the allegedly compatible package versions with versions which we know will work. I believe that this is what you did for #65.

There was [an issue](https://github.com/Snowflake-Labs/schemachange/issues/256) opened in the schemachange repo just before we began running into issues with the incompatible numpy dependency version. But had the package maintainers configured the requirements.txt correctly (e.g., by specifying a numpy version range) then this issue never would have happened. In other words, this was an upstream issue that we found a workaround for.

Proactively pinning the versions of all dependencies (and all their subdependencies and their sub-subdependencies and...?) is not, I think, the best way to go about package management in non-mission-critical situations. It's better to lean on the dependency management system that the Python developers have designed (e.g., PEP 508) and exercise good-faith-until-proven-otherwise that the maintainers of schemachange have thoroughly tested their package in the most common environments.

PR #160, opened 2025-03-04
[SNOW-241] Disable secondary roles by default for all users

philerooski:
See [ticket](https://sagebionetworks.jira.com/browse/SNOW-241) for problem description.

I add a scripting block to `admin/users.sql` which loops over all users in our account, and if their DEFAULT_SECONDARY_ROLES property is set to one of:

* `''` (empty string)
* `<null>`
* `["ALL"]`

then we update the user so that their DEFAULT_SECONDARY_ROLES is set to `[]`. In most cases, this block will only have an effect initially (since default secondary roles is currently enabled for most users) or when we create a new user (in which case Snowflake automatically creates that user with default secondary roles enabled).

### DIFF
--- a/admin/users.sql
+++ b/admin/users.sql
@@ -130,3 +130,36 @@ CREATE USER IF NOT EXISTS ADMIN_SERVICE
 
 CREATE USER IF NOT EXISTS DEVELOPER_SERVICE
     TYPE = SERVICE;
+
+-- Set DEFAULT_SECONDARY_ROLES to [] (do not use secondary roles by default)
+-- for all users
+EXECUTE IMMEDIATE $$
+DECLARE
+    updated_users ARRAY DEFAULT ARRAY_CONSTRUCT(); -- users we updated
+    username STRING; -- a user identifier
+    dsr STRING; -- default secondary role setting
+    user_cursor CURSOR FOR 
+        SELECT "name", "default_secondary_roles" 
+        FROM TABLE(RESULT_SCAN(LAST_QUERY_ID())) 
+        WHERE "name" <> 'SNOWFLAKE'; -- A cursor over our users
+BEGIN
+    SHOW USERS;
+    OPEN user_cursor;
+    LOOP
+        FETCH user_cursor INTO username, dsr;
+
+        -- exit condition
+        IF (username IS NULL) THEN
+            BREAK;
+        END IF;
+
+        -- loop condition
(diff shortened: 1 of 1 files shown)


thomasyu888 (commented):
Nit: My only comment is if there is an account level setting that makes the default to not be enabled, this way this doesn't need to be run every time.

philerooski:
@thomasyu888  

> Nit: My only comment is if there is an account level setting that makes the default to not be enabled, this way this doesn't need to be run every time

Unfortunately, there is not. Basically, Snowflake has set the default value for DEFAULT_SECONDARY_ROLES to ["ALL"] in the CREATE USER command. Alternatively, we can manually specify DEFAULT_SECONDARY_ROLES whenever we create a new user, but I thought this solution would be more seamless / less error-prone.

thomasyu888 (approved):
🔥 LGTM!

admin/users.sql
```diff
@@ -130,3 +130,36 @@ CREATE USER IF NOT EXISTS ADMIN_SERVICE
...
+    LOOP
+        FETCH user_cursor INTO username, dsr;
+
+        -- exit condition
+        IF (username IS NULL) THEN
+            BREAK;
+        END IF;
+
+        -- loop condition
+        IF (dsr = '' OR dsr IS NULL OR dsr = '["ALL"]') THEN
```
jaymedina, line 157: What about `IF dsr != ()` for something more robust? I don't know if `''` '`NULL` or `'["ALL"]'` would always be the only values we should ever look for.

admin/users.sql
```diff
@@ -130,3 +130,36 @@ CREATE USER IF NOT EXISTS ADMIN_SERVICE
...
+            ALTER USER IDENTIFIER(:username) SET DEFAULT_SECONDARY_ROLES=();
+            updated_users := ARRAY_APPEND(updated_users, username);
+        END IF;
+    END LOOP;
+    CLOSE user_cursor;
+    RETURN updated_users;
```
jaymedina, line 163: Is the sole purpose of `updated_users` for logging & making sure we got everybody?

jaymedina (approved):
Just a comment and question, otherwise LGTM!

admin/users.sql
philerooski: Sure, that works too.

PR #164, opened 2025-03-07
[SNOW-248] Add Andrea to `DATA_ANALYTICS` users

jaymedina:
Grant Andrea our COO to Snowflake and DATA_ANALYTICS role

I'm pointing to `dev` from our agreement that everything will flow through `dev` now, even admin commands.


thomasyu888 (approved):
🔥 LGTM!

PR #167, opened 2025-03-14
[SNOW-40] Create certifiedquiz_latest dynamic table (PR 1/2)

rxu17:
### **Problem** 
There came a need for a dynamic table that will reflect the latest state of a record of a user submitted certification quiz on Synapse. Previously, there existed a non-dynamic latest table for the certifiedquiz_latest. This dynamic table contains which user the certified quiz record is for, when it was created, response identifier for the latest submission record of the user and metadata regarding the certified quiz submission (if revoked, if certified, if passed) for the last 14 days.

### **Solution**
A new script is used to introduce the dynamic table. This is [PR #1 of the two part PR](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3794894849/Converting+tables+to+dynamic+tables#Pull-Request-#1) that will integrate in this new dynamic table.

Note that for the column `change_type`, there can be `UPDATE` values outside of `CREATE` values and I've found that this is usually to show a user's certified quiz record being revoked. This is reflected in the column comment.

Note that I reduced the varchar number of characters from original snapshots because they were set to max.

### **Testing**
I followed the [validation steps in this SOP](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3809738755/Validating+a+latest+table+on+synapse_data_warehouse_dev) and tested on my cloned synapse db, generated the dynamic table and compared against the snapshots table in the same cloned db.

1. [Ensure no duplicates](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3809738755/Validating+a+latest+table+on+synapse_data_warehouse_dev#1.-Ensure-no-duplicates)
image

2. [Ensure all records are accounted for](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3809738755/Validating+a+latest+table+on+synapse_data_warehouse_dev#2.-Ensure-all-records-are-accounted-for). 

After making sure that all records are `missing_from_latest` (2a), I ensured that there are no missing records when just comparing the latest which is last 14 days here (2b)
image

3. [Ensure the right records were extracted into *latest](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3809738755/Validating+a+latest+table+on+synapse_data_warehouse_dev#3.-Ensure-the-right-records-were-extracted-into-*latest)

Dynamic Table
image
Snapshots Table
image
It's a match! Did this with some more user_ids

4. [Snowsight check](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3809738755/Validating+a+latest+table+on+synapse_data_warehouse_dev#4.-Snowsight-check)

Looks as expected
image


synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
thomasyu888, line 1: This looks good, although I think we can as a team discuss the differences between 

* certifiedquiz_latest: This containing the latest information per user of a certification quiz
* certifiedquiz_deduplicated: this contains all quiz responses (partitioned by ID)

Thats for another PR, but the reason I ask this is because the metrics you would gather from the two tables above is different.  the de-duplicated table can answer questions like: how many times did a USER take a quiz before they passed, whereas the latest answers the question around what is the latest state of certified quiz per user.

thomasyu888 (approved):
🔥 LGTM.  I'll defer to @philerooski / @jaymedina for final review of this two part PR.

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
```diff
@@ -0,0 +1,50 @@
+USE SCHEMA {{database_name}}.SYNAPSE; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE DYNAMIC TABLE CERTIFIEDQUIZ_LATEST_DYNAMIC
```
philerooski, line 3: I assume the name will be changed to match the formatting of the other tables in part 2?

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
```diff
@@ -0,0 +1,50 @@
...
+                REVOKED,
+                REVOKED_ON,
+                CERTIFIED
+            FROM
+                {{database_name}}.SYNAPSE_RAW.CERTIFIEDQUIZSNAPSHOTS --noqa: TMP
+            WHERE
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '14 DAYS'
+            QUALIFY ROW_NUMBER() OVER (
+                    PARTITION BY USER_ID
+                    ORDER BY INSTANCE DESC, RESPONSE_ID DESC
```
philerooski, line 44: This ordering could be correct, but doesn't it depend on the RESPONSE_ID incrementing if a user takes a certification quiz more than once on the same INSTANCE (stack version)? If we want the latest certification quiz, does it make more sense to order by CHANGE_TIMESTAMP DESC and SNAPSHOT_TIMESTAMP DESC? This would match the process we use in many of our other tables.

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
rxu17: Yes according to the SOP we will drop the `_DYNAMIC` part in PR 2/2

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
philerooski: As Tom and I discussed in our tête-à-tête, the current implementation is correct for a certifiedquiz_latest table.

certifiedquiz_deduplicated is a new class of what are traditionally called _fact tables_, and which actually more closely approximate a silver-layer table than our *latest tables. We will be providing this table eventually, just not as part of the work on this ticket.

rxu17: Do we technically have any _fact tables_ current amongst our tables?

thomasyu888: @philerooski / @rxu17 , Taking a step back after the 1 on 1, I thought about this a little bit. I think it's worthwhile to do a tiny bit of research on what fact tables vs record/log tables are.

On a high level
* Fact Tables: Store measurable business data (e.g., sales, revenue, counts) and are typically structured around a set of dimensions. Fact tables contain numeric values that can be aggregated. Example: A table storing daily sales transactions (sales_fact), with columns like date, store_id, product_id, and sales_amount.
* Record Tables: Store raw, transactional, or log-like data without aggregation or summarization. These tables often serve as an audit trail and may not fit neatly into a dimensional model. Example: A table storing API request logs, where each row represents a single API request, including timestamp, endpoint, response_time, and status_code.

So I would actually say `fileupload`, `filedownload`, `processedaccessrecords`, `certifiedquiz (records)` are more like record/ log tables, but I'm open to more discussion.

philerooski: > Do we technically have any fact tables current amongst our tables?

@rxu17 Yes. Anything that has a record per event (e.g., processedaccess, filedownload, certifiedquiz) is a transactional fact table. Everything @thomasyu888 is definitely a (subtype of) fact table. Wikipedia has a decent way of explaining the types of fact tables: https://en.wikipedia.org/wiki/Fact_table#Types_of_fact_tables

There is some gray area when it comes to tables like FILEINVENTORY. If those snapshots capture changes in S3 objects (which I don't think would ever happen?), then it behaves like a fact table. Otherwise it behaves more like a dimension table.

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
rxu17: Thanks for the question. I wanted to preserve the original latest non-dynamic table filtering and found that the validation succeeded, but after thinking about it further and exploring ([see investigation results here](https://sagebionetworks.jira.com/browse/SNOW-40?focusedCommentId=243010)), I was able to answer both questions

**Summary**
- There are some violations of using my current ordering to get the latest submission (even though technically we get the latest submission but we don't get the latest timestamp values in that record)
- We do need to order by `RESPONSE_ID` in addition to the `CHANGE_TIMESTAMP` and `SNAPSHOT_TIMESTAMP` columns because there can be two responses for the same `CHANGE_TIMESTAMP` and `SNAPSHOT_TIMESTAMP`

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
BryanFauble: Is that other PR made yet? Wondering if we would be able to just modify this PR if not

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
```diff
@@ -0,0 +1,50 @@
...
+                REVOKED_ON,
+                CERTIFIED
+            FROM
+                {{database_name}}.SYNAPSE_RAW.CERTIFIEDQUIZSNAPSHOTS --noqa: TMP
+            WHERE
+                SNAPSHOT_TIMESTAMP >= CURRENT_TIMESTAMP - INTERVAL '14 DAYS'
+            QUALIFY ROW_NUMBER() OVER (
+                    PARTITION BY USER_ID
+                    ORDER BY RESPONSE_ID DESC, CHANGE_TIMESTAMP DESC, SNAPSHOT_TIMESTAMP DESC
+                ) = 1
```
BryanFauble, line 45: Q1: Could it be possible for the same `USER_ID` to take the quiz multiple times, and if so, would we want to now about all those attempts?

Q2: Should this also take into account the `STACK` as well? It's a bit of an edge case specific to those with a dev/staging account, however, I wonder if this could cause us to drop a record from the prod stack if someone was doing testing in staging.

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
rxu17: Nope it's not. Hmm what do you mean by modify this PR?

BryanFauble: > `we will drop the _DYNAMIC part in PR 2/2`


Could this just be done in this PR 1/2? Is there a technical problem here that prevents us from making this change?

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
rxu17: Going to let @thomasyu888 and @philerooski chime in here for these to add any extra context

Q1: There was discussion earlier, and if I'm understanding it correctly, the conclusion of that discussion is that this table should just be the latest state table (so latest certified submission quiz state for a user) and we should have another de-duplicated table of all the latest submissions that a single user will take.

Q2: Great question, maybe I don't understand what `STACK` means then? We have `dev` and `prod` databases, I thought the `STACK` presents what database we are in so we'd have separate `certified quiz snapshots` tables for each database so there wouldn't be a snapshots table that has both `prod` and `dev` when we pull from it to create this table?

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
rxu17: Trying to go back to old comments on the [SOP for adding dynamic table](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3794894849/Converting+tables+to+dynamic+tables) but I think the reasoning behind doing this instead of before which was having the new table, stream and task dropping all in one PR / another SOP revision before this one was renaming the old table to _BACKUP and manually updating all the streams/tasks to use it):
- So we can validate the dynamic table first while preserving the original table (so the streams and tasks still using the old table doesn't need to be manually updated to use the new table / if anything is wrong with table it doesn't breaking their workflows)
- Once we merge in this one, and verify everything is as expected, we can do all the stream and task dropping in the second PR

philerooski: Yeah, the gist is that we can compare a table which is being actively updated via the old stream/task mechanism to the new dynamic table. The naming/renaming will only happen in the dev database while we validate this ticket.

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
BryanFauble: > certified quiz snapshots tables for each database so there wouldn't be a snapshots table that has both prod and dev when we pull from it to create this table.

Great, if that's already filtered out - then there isn't a concern here

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
philerooski: > ...because there can be two responses for the same CHANGE_TIMESTAMP and SNAPSHOT_TIMESTAMP...

Huh, what does this mean from like a Synapse user perspective? A user submits a certification quiz and somehow two `RESPONSE_ID`s are created at the same time? Am I understanding this correctly?

BryanFauble (approved):
Nothing looks out of place to me!

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
rxu17: I thought 1 submission of a certification quiz = 1 response_id created? 
So some users have multiple `RESPONSE_ID`s because they submit it multiple times.

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
philerooski: I looked at the distinct values of `stack` and in SYNAPSE_DATA_WAREHOUSE this column is only ever "prod" and in SYNAPSE_DATA_WAREHOUSE_DEV it's only ever "dev".

synapse_data_warehouse/synapse/dynamic_tables/V2.46.0__certifiedquiz_latest_dynamic_table.sql
philerooski: > I thought 1 submission of a certification quiz = 1 response_id created?

This is also what I thought
> So some users have multiple RESPONSE_IDs because they submit it multiple times.

But these ought to have different CHANGE_TIMESTAMP, no?

I looked at the data:
```
select change_timestamp, snapshot_timestamp, response_id, count(*) as n
from synapse_data_warehouse.synapse_raw.certifiedquizsnapshots
group by change_timestamp, snapshot_timestamp, response_id
having n > 1
order by n desc;
```
And there are duplicates when we look at just change_timestamp, snapshot_timestamp, and response_id, so maybe we do need to consider the response_id when deduplicating...

BUT these records appear to be the same records in this query, which means they are true duplicates: every value is the same.
```
select *, count(*) as n
from synapse_data_warehouse.synapse_raw.certifiedquizsnapshots
group by all
having n > 1
order by n desc;
```

btw, I also did this query in SYNAPSE_DATA_WAREHOUSE_DEV, and there was a difference of 1 (114 vs 113) records between these two queries, but I can't think of why that would be the case.

rxu17: > But these ought to have different CHANGE_TIMESTAMP, no?

So that was something I came across. That there were instances where the initial response_id had a different change_timestamp than the same response_id for later records. Then at some point it had the same `CHANGE_TIMESTAMP` as another response_id. See the first [**conclusion**]( https://sagebionetworks.jira.com/browse/SNOW-40?focusedCommentId=243010) comment where I had an example for a user.

philerooski: @rxu17 This is also what I found in my previous comment. But in the case of prod, these were true duplicates: every value was the same. In the case of dev there was one record which had duplicate change_timestamp/snapshot_timestamp/response_id values, but at least one of the other fields was different. I'm not sure what the discrepancy is.

rxu17: Based on discussion with Phil, going to file an IT ticket for the above (seeing different `CHANGE_TIMESTAMP` values for the same `RESPONSE_ID` and `CHANGE_TYPE`). Otherwise, it has been deemed good to merge!

PR #168, opened 2025-03-19
Add finance users

thomasyu888:
# Problem
Finance team need to be added and people who have left sage can be removed.

# Solution
Added the finance team

# Test
NA

This is an exception to fast track adding users to PR into `main`

### DIFF
--- a/admin/users.sql
+++ b/admin/users.sql
@@ -22,7 +22,6 @@ CREATE USER IF NOT EXISTS "xindi.guo@sagebase.org";
 CREATE USER IF NOT EXISTS "alexander.paynter@sagebase.org";
 CREATE USER IF NOT EXISTS "angie.bowen@sagebase.org";
 CREATE USER IF NOT EXISTS "ashley.clayton@sagebase.org";
-CREATE USER IF NOT EXISTS "lakaija.johnson@sagebase.org";
 CREATE USER IF NOT EXISTS "amber.nelson@sagebase.org";
 CREATE USER IF NOT EXISTS "tiara.adams@sagebase.org";
 CREATE USER IF NOT EXISTS "jessica.vera@sagebase.org";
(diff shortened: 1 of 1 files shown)


jaymedina (approved):

BryanFauble (approved):

admin/users.sql
```diff
@@ -127,7 +124,7 @@ CREATE USER IF NOT EXISTS RECOVER_SERVICE
     MUST_CHANGE_PASSWORD = TRUE;
 
 CREATE USER IF NOT EXISTS ADMIN_SERVICE
-    TYPE = SERVICE;
+    TYPE = SERVICE; --noqa: LT02, PRS
```
philerooski, line 127: Do we need these rule exceptions? The indent is already a multiple of 4 (LT02) and (PRS) doesn't seem to exist: https://docs.sqlfluff.com/en/stable/reference/rules.html#rule-layout.indent

philerooski (changes requested):
Removing the line which creates the user does not remove the user. You need to add a `DROP USER IF EXISTS <name>` line for each of these.

admin/users.sql
philerooski: Ohh, PRS is unparseable section. Is it just me, or has sqlfluff been incredibly buggy? It even considers `DROP DATABASE ROLE` statements unparseable, even with dialect set to Snowflake.

thomasyu888:
> Removing the line which creates the user does not remove the user. You need to add a `DROP USER IF EXISTS <name>` line for each of these.

@philerooski I haven't been removing the user, instead i've been manually "suspending" their account in snowflake just in case their accounts are tied to roles and resources they own.

admin/users.sql
thomasyu888: Ya... I think `sqlfluff` is definitely buggy.. I think it just takes time to add features.

philerooski:
@thomasyu888 Do you want to
```
ALTER USER janesmith SET DISABLED = TRUE;
```
Or just keep it casual?

thomasyu888:
> ```
> ALTER USER janesmith SET DISABLED = TRUE;
> ```
> 
> Or just keep it casual?

@philerooski I think i probably should've had a V script that disabled them, but at this point, all of those users have all been disabled manually from the UI. 

The nice part about using SSO is that people leaving sage lose access to their sage emails, which means they won't be able to access snowflake anyhow.

philerooski (approved):

PR #170, opened 2025-03-21
[DPE-1275] Create a snapshot_latest dynamic table

BryanFauble:
# Problem

Ticket: [DPE-1275](https://sagebionetworks.jira.com/browse/DPE-1275)

Problem is described in this slack comment here: https://sagebionetworks.slack.com/archives/C02M7QR8D46/p1742578856494009

TLDR:
`node_latest` does not contain enough information to get the state of an entity at a specific version.

# Solution

1. Create a dynamic table based off the DDL for the node_latest table, but including version_number

# Testing

 

[DPE-1275]: https://sagebionetworks.jira.com/browse/DPE-1275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/synapse_data_warehouse/synapse/dynamic_tables/V2.44.0__snapshot_latest.sql
+++ b/synapse_data_warehouse/synapse/dynamic_tables/V2.44.0__snapshot_latest.sql
@@ -1,0 +1,61 @@
+USE SCHEMA {{database_name}}.synapse; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE DYNAMIC TABLE SNAPSHOT_LATEST
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
+    AS
+        WITH latest_unique_rows AS (
+            SELECT
+                CHANGE_TYPE,
+                CHANGE_TIMESTAMP,
+                CHANGE_USER_ID,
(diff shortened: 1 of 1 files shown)


philerooski:
It's a small change to the existing NODE_LATEST table, but conceptually it's somewhat radical. Up to this point, we've had a one-to-one relationship between our *raw and *latest data types. This bucks that model by giving us a second table derived from NODESNAPSHOTS. It also uses a *latest suffix but is qualitatively different from what we've done with our other *latest tables (which only reflect the latest state of Synapse, or in this case, the latest version of a node). 

One question I have is whether this implementation is general enough, or whether it might be too specific to your use-case. When we build out "silver-layer" historical data tables, will this table become largely redundant?

Anyway -- the short of it is that it might be premature at this point to officially support a table like this for other analysts to use, and you may be better off building this as part of your own pipeline (at least, until we can put together a version that we feel confident can cover a broad array of use-cases for historical NODE data).

BryanFauble:
Closing this PR out. The plan will instead be to create an `experimental` dynamic table with the required data. And use that for this purpose.

PR #173, opened 2025-03-21
Create new warehouse

thomasyu888:
# Problem

The `compute_xsmall` warehouse was being flooded with auto DYNAMIC TABLE refreshes. https://sagebionetworks.slack.com/archives/C05PBT2FUHF/p1742586409640229
> NOTE, this is a direct push to prod

# Solution

 1. Create a solution so that the data analysts across sage doesn't have to compete with the resources.
There doesn't seem to be a serverless option for DYNAMIC TABLES

# Testing

NA

PR #174, opened 2025-03-21
Set warehouse to be dpe_automation

thomasyu888:
# Problem

The dynamic tables are using `compute_xsmall` right now which can eat up a lot of query time for people using snowflake

# Solution

Alter all dynamic tables to use a specific automation warehouse so our users don't have to compete with resources.

# Testing

NA

 

depends on #173


rxu17 (approved):
LGTM!

BryanFauble:
does this need to be updated as well?

1. https://github.com/Sage-Bionetworks/snowflake/blob/4eca67aae11f624f844b2359268b4fdc6cf2fa3f/.github/actions/configure-snowflake-cli/action.yml#L19

synapse_data_warehouse/synapse/dynamic_tables/V2.44.0__alter_all_dynamic_tables_to_use_different_warehouse.sql
thomasyu888, line 1: Need to check if this is the right schemachange version script.

philerooski:
Could you go into more detail about why you think creating a new warehouse is the right solution for managing concurrent queries across the org? This seems like the exact situation that [multi-cluster warehouses](https://docs.snowflake.com/en/user-guide/warehouses-multicluster#what-is-a-multi-cluster-warehouse) were designed to address.

> By default, a virtual warehouse consists of a single cluster of compute resources available to the warehouse for executing queries. As queries are submitted to a warehouse, the warehouse allocates resources to each query and begins executing the queries. If sufficient resources are not available to execute all the queries submitted to the warehouse, Snowflake queues the additional queries until the necessary resources become available.

> With multi-cluster warehouses, Snowflake supports allocating, either statically or dynamically, additional clusters to make a larger pool of compute resources available.

thomasyu888:
@philerooski 

> COMPUTE_XSMALL is currently configured with a MAX_CLUSTER_COUNT of 1, but we can dynamically increase the number of provisioned clusters to 300 if we really wanted to.

That is a great point, and I think we can still set the max_cluster_count to be higher than 1 for compute_xmall when more users across the org use it. however, there is another use case.  I'd like to know how much compute credits all of our automation is using just for tracking purposes.

image

I'm not sure how you could achieve that with the multi-cluster warehouse.

philerooski:
> however, there is another use case

Let's create a ticket for this use case and explore these use cases more generally (Are there other scenarios we would like cost tracking?). I'm a little wary of trying to swap out the infrastructure a big chuck of our data warehouse pipeline runs on in one shot the Friday before home week.

I'm okay with increasing the max cluster count of COMPUTE_XSMALL in the meanwhile.

thomasyu888:
> > however, there is another use case
> 
> Let's create a ticket for this use case and explore these use cases more generally (Are there other scenarios we would like cost tracking?). I'm a little wary of trying to swap out the infrastructure a big chuck of our data warehouse pipeline runs on in one shot the Friday before home week.
> 
> I'm okay with increasing the max cluster count of COMPUTE_XSMALL in the meanwhile.

@philerooski , makes sense, i like that idea.  Could you help me alter the compute_xsmall in the meantime?

philerooski:
@thomasyu888 
> Could you help me alter the compute_xsmall in the meantime?

I'll open a PR

thomasyu888:
Closing this PR to have a more in-depth discussion of separating out the compute for automation vs users

PR #177, opened 2025-03-21
Rewrite grants to proxy admins to reference current name

philerooski:
# Problem

The previous PR #176 didn't fix the CI, since both the proxy admin roles with the former name and the proxy admin roles with the current name already exist.

# Solution

I figured the least pedantic way to resolve this would be to rewrite the original `CREATE ROLE IF NOT EXISTS...` statements in `admin/roles.sql` to reference the current name and update the references in `admin/grants.sql`. I decided not to update the references in the versioned scripts since we'll end up getting a warning in perpetuity that the hash of these scripts has changed, which I think is more confusing then leaving them alone. This brings to light some of the challenges with tracking changes when objects change names.

I will drop the former roles (`SYNAPSE_DATA_WAREHOUSE_ALL_ADMIN` and `SYNAPSE_DATA_WAREHOUSE_DEV_ALL_ADMIN`) manually behind the scenes.

# Testing


thomasyu888 (approved):
LGTM!

PR #178, opened 2025-03-21
Remove snowsql variable substitue indicator

philerooski:
# Problem

The [previous CI run](https://github.com/Sage-Bionetworks/snowflake/actions/runs/14001491450/job/39208679209) failed because `admin/integrations.sql` contained a `snowsql` client-specific command which was formerly necessary for variable substitution.

# Solution

The new Snowflake CLI natively supports variable substitution, so I removed this command.

# Testing

### DIFF
--- a/admin/integrations.sql
+++ b/admin/integrations.sql
@@ -1,5 +1,3 @@
-!set variable_substitution=true; --noqa: PRS
-
 USE ROLE ACCOUNTADMIN;
 
 -- * Integration to prod (SNOW-14)


thomasyu888 (approved):
I just want to be certain that this "GOOGLE_SSO" doesn't get overwritten, but otherwise, I approve! 

It would be not great to deal with google sso being down on Friday

philerooski:
@thomasyu888 Which parts worry you? I'm thinking that...

1. We already invoke this like `create security integration IF NOT EXISTS GOOGLE_SSO...`, so this line shouldn't even be run, since GOOGLE_SSO already exists.
2. If the line were to run, the variables -- which I had to change the format of to match the format used by the Snowflake CLI -- are all related to authentication. If those were broken somehow then we just wouldn't be able to authenticate, I think?

thomasyu888:
@philerooski good enough for me! Thanks!

philerooski:
@thomasyu888 I dug a little deeper and found that [the last time](https://github.com/Sage-Bionetworks/snowflake/actions/runs/13997687857/job/39196815094) we successfully ran admin/integrations.sql (while deploying 6a3f23d), the GOOGLE SSO command had a syntax error:
```
001008 (22023): SQL compilation error:
invalid value ['&saml2_sso_url'] for parameter 'saml2_sso_url'
```
which GPT warned me might be an indication that we need to double quote the environment variables when we pass them to the Snowflake CLI in ci.yaml. If this is a URL that makes sense, so good thing you brought it up!

philerooski:
That said, the GOOGLE_SSO integration exists, so this command won't be run even after we fix the way we're passing env variables.

BryanFauble (approved):

PR #181, opened 2025-03-21
Increase max cluster count for compute_xsmall

philerooski:
# Problem

We recently encountered some issues where queries were queueing because of capacity constraints in our default warehouse (COMPUTE_XSMALL).

# Solution

I increase the max cluster count from 1 to 10. At first, a ten-fold increase in maximal compute resources might seem dramatic, although this is [the default](https://docs.snowflake.com/en/user-guide/warehouses-multicluster#upper-limit-on-number-of-clusters-for-a-multi-cluster-warehouse) for warehouses. Since our min cluster count is still 1, our warehouse cluster count increases dynamically ([docs](https://docs.snowflake.com/en/user-guide/warehouses-multicluster#upper-limit-on-number-of-clusters-for-a-multi-cluster-warehouse)). Since our max cluster count is 10 or less and we use the standard scaling policy, Snowflake will start one additional cluster if "a query is queued, or if Snowflake estimates the currently running clusters don’t have enough resources to handle any additional queries" ([docs](https://docs.snowflake.com/en/user-guide/warehouses-multicluster#setting-the-scaling-policy-for-a-multi-cluster-warehouse)).

# Testing


admin/warehouses/V1.11.0__increase_max_cluster_count.sql
rxu17, line 1: Does this mean we could technically have ~200 concurrent queries given our max query right now for 1 cluster size warehouse is ~20?

BryanFauble: What was a bit odd was that earlier when queries were getting queued up was that there was only 2 or 3 queries actively running.

I wonder if there is a "Number of queries" hard limit of 20, and a soft limit when the queries that are being executed consume X amount of resources.

BryanFauble (approved):
LGTM

admin/warehouses/V1.11.0__increase_max_cluster_count.sql
philerooski: > I wonder if there is a "Number of queries" hard limit of 20

fwiw, I couldn't find any evidence for this. The [docs](https://docs.snowflake.com/en/user-guide/warehouses-overview#query-processing-and-concurrency) seem to suggest it's more complicated than that:
> The number of queries that a warehouse can concurrently process is determined by the size and complexity of each query. As queries are submitted, the warehouse calculates and reserves the compute resources needed to process each query. If the warehouse does not have enough remaining resources to process a query, the query is queued, pending resources that become available as other running queries complete.

rxu17 (approved):
LGTM!

thomasyu888 (approved):
Thanks! Let's monitor the cost in the next month to make sure it doesn't go overboard.

PR #193, opened 2025-04-11
Update CONTRIBUTING.md

jaymedina:
# Problem

Ticket: [SNOW-90](https://sagebionetworks.jira.com/browse/SNOW-90)

There was some subtle wording that needed to be modified in the `CONTRIBUTING.md` after switching to our opt-out system, and the reference to `drop_clone` needed to be removed.

# Solution

-  [x] Update the documentation accordingly

# Testing

N/A

 

[SNOW-90]: https://sagebionetworks.jira.com/browse/SNOW-90?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,34 +56,26 @@ Once you've made your changes and committed them locally, push your branch to th
 git push origin snow-123-new-feature
 ```
 
-### 5. Create a Draft Pull Request
+### 5. Open a Draft Pull Request
 
-In order to initiate automated testing you will need to work on a draft pull request (PR) on GitHub. After pushing your commits to
-the remote branch in Step 4, use the GitHub UI to initate a PR and convert it to draft mode.
+After pushing your commits in Step 4 to your remote branch, open a pull request (PR) against the dev branch using the GitHub UI. Start your PR in draft mode, and mark it as ready for review after you've implemented all your necessary changes.
 
-After testing your changes against `schemachange` using the instructions in [Running CI Jobs for Database Testing](#running-ci-jobs-for-database-testing),
-you can then take your PR out of draft mode by marking it as Ready for Review in the GitHub UI.
+Opening your PR (whether in draft or not) will trigger the automated test workflows described in the next section. We recommend reading that section to understand how to further test your changes, OR if you are not introducing changes to the actual Snowflake data repository and want a way to opt-out of these workflow runs.
 
 ## Running CI Jobs for Database Testing
 
-This repository includes automated CI jobs to validate changes against a cloned database. If you want to trigger these jobs to test your changes in an isolated database environment on Snowflake, please follow the steps below:
+This repository includes automated CI jobs designed to validate your changes against a cloned version of the database in an isolated environment on Snowflake. Please follow the instructions below to enable this testing:
 
 ----
 
-### 1. Make A Commit
+### 1. Open a Draft Pull Request (if not done already)
 
-By default, each new commit you make in a PR once you have initialized it will trigger the `test_with_clone` job for your branch. This job does two things:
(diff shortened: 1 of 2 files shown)


jaymedina:
@philerooski I kind of want to remove the trigger for opened/reopened PRs and just have the schemachange job trigger for new commits in an existing PR, because this doesn't give you the chance to opt-out from the get-go. I was able to interrupt it here by making another commit right after labeling with `skip_clone`

philerooski:
@jaymedina I was able to open a PR without triggering a clone by adding the label before creating the pull request here: https://github.com/Sage-Bionetworks/snowflake/pull/194

CONTRIBUTING.md
```diff
@@ -96,15 +100,28 @@ You can go on Snowsight to perform manual inspection of the changes to your sche
...
+1. Log onto Snowsight (the Snowflake web client)
+2. Navigate to `Databases` on the sidebar:
+
+   <img width="200" alt="image" src="https://github.com/user-attachments/assets/076f6a68-69e6-4072-8562-753769be1a93" />
+
+3. Navigate to your clone in the new panel, and hit `Drop` from the `...` dropdown:
+
+   <img width="939" alt="image" src="https://github.com/user-attachments/assets/3155b26b-ced0-44f6-8563-6da4daf9cf0e" />
+
+
```
philerooski, line 124: Does this work? The clone will be owned by `<DOMAIN>_ADMIN`, which I'm not sure at this point whether we will grant this role to developers.

jaymedina: Yeah, this doesn't work unless you're <DOMAIN>_ADMIN. Unless you want to spend a little more time on the implementation design, I can rescope this PR to introduce a new PR label for manually triggering dropping the database (`drop_clone`). Otherwise I can make a ticket so we can think about this a little bit more, and remove this section entirely for now.

philerooski: `drop_clone` is an option. We could have a ruleset that gives priority to this label, so as long as its present the `test_with_clone` job doesn't run and the `drop_clone` job does run. Although I think it's a good rule of thumb that as long as a PR is open, the corresponding clone should be available for reviewers to poke around in. Closing a PR is a simple way to say "These changes are no longer open for review and the clone should be dropped".

Did you have a specific use-case in mind for dropping the clone while the PR is still open?

jaymedina: > Did you have a specific use-case in mind for dropping the clone while the PR is still open?

No, this section was mainly to minimize cost, but we haven't had any cost issues from our current clone databases, so I'm leaning towards scrapping this section entirely for now.

> Although I think it's a good rule of thumb that as long as a PR is open, the corresponding clone should be available for reviewers to poke around in.

Agreed

jaymedina: LMK what you think of [these edits](https://github.com/Sage-Bionetworks/snowflake/pull/193/commits/f2726602db81af2feca8ba863a700c7fbd93add1)

philerooski:
> 5. Create a Draft Pull Request

Maybe a nit, but this seems like it should be step 1 of "Running CI Jobs for Database Testing", not the final step in "Getting started". Or perhaps we could split this step among both sections. "Getting started" step 5 could briefly state that we need to open a draft PR into dev, and step 1 of "Running CI Jobs for Database Testing" could then discuss the test_with_clone workflow that was triggered by this PR and the "After testing your changes against schemachange..." part.

> 1. Make a commit

This then makes more sense as step 2 of "Running CI Jobs for Database Testing", since the initial step in running the test_with_clone workflow is actually to open a PR.

philerooski (approved):

PR #191, opened 2025-04-09
[SNOW-266] SYNAPSE_AGGREGATE schema

philerooski:
# Problem

Ticket: [SNOW-266](https://sagebionetworks.jira.com/browse/SNOW-266)

# Solution

1. Create gold layer schema (synapse_aggregates)
    - See 4511b9d
2. Ensure all the database roles are created (similar access to silver schema)
    - See 3565f13
    - I have created admin, analyst, and developer-specific database roles and granted ownership and usage to the relevant roles.
5. Create a dummy table to show that schemachange can push data into this schema (ensure CI/CD works) - maybe project ToS? https://sagebionetworks.slack.com/archives/C05PBT2FUHF/p1738958039708949?thread_ts=1738947601.513069&cid=C05PBT2FUHF
    - See cabd83f
7. Ensure that the service user can query across silver and bronze table since aggregates may need tables from both schemas
    - The role used to run schemachange is either SYNAPSE_DATA_WAREHOUSE_DEV_ADMIN or SYNAPSE_DATA_WAREHOUSE_ADMIN, both of which inherit ownership of all objects in their respective database.

This PR provides an example of how our RBAC framework can be used to manage privileges on a new schema and its objects. Of particular note is:

* There is a database role per archetype (admin/analyst/developer)
* Each of these database roles is inherited by the archetypical account role
* Ownership of all database roles is consolidated with the admin database role (SYNAPSE_AGGREGATE_ALL_ADMIN), except for the admin database role itself, which is owned by the (proxy) admin account role (SYNAPSE_DATA_WAREHOUSE_PROXY_ADMIN)

This is illustrated in the diagram below:
 
SNOW-266

# Testing

I manually tested the changes in `synapse_data_warehouse/database_roles/V2.47.1__synapse_aggregate_database_roles.sql` (essentially -> constucting the diagram above) in my own test database. I also created future grants and created a table which I then queried using the analyst role. I provide sample code below.

<details><summary>Details</summary>
<p>

```
USE ROLE SYSADMIN;
USE DATABASE SYNAPSE_DATA_WAREHOUSE_PHIL;

-- Create schema
CREATE OR REPLACE SCHEMA SYNAPSE_AGGREGATE;

-- Create "admin" database role which will own the synapse_aggregate objects
CREATE OR REPLACE DATABASE ROLE SYNAPSE_AGGREGATE_ALL_ADMIN;

-- Grant ownership of this database role to the proxy admin
GRANT OWNERSHIP
[...]

PR #195, opened 2025-04-14
Drop tables used for testing

philerooski:
# Problem

There is no ticket for this PR.

We want to make a release to prod/main. Some of the commits currently in dev were left in an intermediary state so that we could test them. It could be confusing to deploy these intermediate states.

# Solution

This PR drops the test resources in anticipation of doing a regular release to prod/main. I've specified which commit and resources these are below:

- [c8a58f1](https://github.com/Sage-Bionetworks/snowflake/commit/10ac0de3b42fc91e1ee97e68e0176abeec8a58f1) created a CERTIFIEDQUIZ_LATEST_DYNAMIC table
- [7c83d05](https://github.com/Sage-Bionetworks/snowflake/commit/78d05ffc975ad3d6984af1c983a69b49b7c83d05) created a DUMMY_TABLE table.

# Testing


philerooski:
@jaymedina I noticed that DUMMY_TABLE in the SYNAPSE_AGGREGATE schema was unable to be dropped in the clone. I think this is because there should be a line in the "Transfer ownership of current and future tasks and dynamic tables to `<CLONE>_PROXY_ADMIN`" step where we transfer ownership of the SYNAPSE_AGGREGATE_ALL_ADMIN database role to the `<CLONE>_PROXY_ADMIN`.

I made this change as part of the PR. These changes won't take effect until we merge this PR into dev. LMK what you think, and if you can think of any other changes we might need to make to the workflow to account for the new SYNAPSE_AGGREGATE schema.

.github/workflows/test_with_clone.yaml
```diff
@@ -150,7 +150,7 @@ jobs:
                 snow sql -q "USE ROLE SECURITYADMIN; GRANT OWNERSHIP ON FUTURE DYNAMIC TABLES IN SCHEMA ${SNOWFLAKE_DOMAIN_CLONE_SANITIZED}.SYNAPSE TO ROLE ${SNOWFLAKE_CLONE_PROXY_ADMIN_ROLE};"
                 snow sql -q "USE ROLE SECURITYADMIN; GRANT OWNERSHIP ON FUTURE TASKS IN SCHEMA ${SNOWFLAKE_DOMAIN_CLONE_SANITIZED}.SYNAPSE_RAW TO ROLE ${SNOWFLAKE_CLONE_PROXY_ADMIN_ROLE} COPY CURRENT GRANTS;"
```
jaymedina, line 151: Add one more query for `synapse_aggregate` schema here

jaymedina:
For some reason the dummy table is owned by <DOMAIN>_ADMIN, rather than <DOMAIN>.<SCHEMA>_ALL_ADMIN, and I think that's the reason this is still failing.

jaymedina:
Here is a breakdown of the problem as I'm seeing it:

* You correctly grant ownership of all tables to `SYNAPSE_AGGREGATE_ALL_ADMIN` in the `admin/grants.sql` of [this PR](https://github.com/Sage-Bionetworks/snowflake/pull/191/files#diff-d23e4b75e1cfeaf9ee3bf2f274210e62505eb30c14fdc86f985cf625d04de928)
* This change doesn't take effect until a push to `prod` which is why `<DOMAIN>_ADMIN` owns the dummy table
* Because `<DOMAIN>_ADMIN` owns the dummy table, `<CLONE>_PROXY_ADMIN` is never able to have USAGE of this table, because it has USAGE of the table by being granted the `<SCHEMA>_ALL_ADMIN` database roles.

jaymedina (approved):
🔧 Approved pending closed threads and a response to my comment. Thanks!

thomasyu888 (approved):
🔥 LGTM!

.github/workflows/test_with_clone.yaml
philerooski: There aren't any tasks or dynamic tables in SYNAPSE_AGGREGATE. Would it be premature to add a line here?

jaymedina: No, since the logic in the PR for `synapse_aggregate` implies we will be introducing dynamic tables we should prepare for that in the CI/CD. We _could_ opt-out the FUTURE TASKS query, though, if you're not planning to have tasks in the gold-table schemas.

philerooski: Makes sense. I've added the line for DYNAMIC TABLES.

philerooski: Whoops, I originally misread this. I added two lines: one for granting ownership of all dynamic tables to the clone proxy admin (even though we don't yet have any dynamic tables) and another for revoking ownership of future dynamic tables from the `<DOMAIN>_PROXY_ADMIN`

jaymedina: lgtm!

PR #197, opened 2025-04-15
Move synapse_aggregate ownership grants

philerooski:
# Problem

There is no ticket for this PR. I noticed that there was [an issue](https://github.com/Sage-Bionetworks/snowflake/actions/runs/14458334361/job/40546212649) running `admin/grants.sql` during the deployment of #196. This was caused by an OWNERSHIP grant being attempted without the REVOKE/COPY CURRENT GRANTS clause on an object which already had privileges upon it. 

Actually, it's a mystery to me why DATA_ENGINEER was granted CREATE DYNAMIC TABLE privileges on the SYNAPSE_AGGREGATE schema. It seems like it originated from [a former grant on future schemas](https://github.com/Sage-Bionetworks/snowflake/blob/382086a12853dbafaae502f280b302fb112ea05e/admin/grants.sql#L250-L261) in the SYNAPSE_DATA_WAREHOUSE database. This future grant has already been revoked but I guess the Snowflake backend wasn't as fast as my CI/CD.

# Solution

I moved ownership grants to a versioned file and removed them from `admin/grants.sql`. I revoke current grants, but any grants besides OWNERSHIP which ought to be applied will be reapplied in `admin/grants.sql`.

# Testing


thomasyu888 (approved):
🔥 LGTM! I have noticed Snowflake to be slow on grants sometimes as well...

philerooski:
@thomasyu888 Thanks. I'll pull this into main as well and include it with the release. I've already tested locally so should be no further issues.

PR #200, opened 2025-04-29
Update grants.sql

jaymedina:
# Problem

Jess Britton is not on Snowflake but needs access to dashboards

# Solution

add Jess

# Testing


BryanFauble (approved):
LGTM

PR #201, opened 2025-05-01
[SNOW-258] Introduce O1KR1 Aggregate Table

jaymedina:
# Problem

Ticket: [SNOW-258](https://sagebionetworks.jira.com/browse/SNOW-258)

We need tables to more-readily provide metrics and insight on our company-wide OKR progress. Namely, we want to create a table that will provide insight on the following OKR:

**_O1.KR1: Increase number of users who contribute data on Synapse and portals by 10% from 2024 to 2025_**

# Solution

We are now introducing the "gold" layer of our medallion architecture in Snowflake which includes aggregate tables across the time dimension. The following data model serves to highlight the structure of our first aggregate table:

** THIS DIAGRAM IS OUTDATED **
Editor _ Mermaid Chart-2025-04-30-184953

**Out-of-scope**

- [x] Fix a typo in granting ownership of future dynamic tables to the clone proxy admin role

# Testing

[SNOW-258]: https://sagebionetworks.jira.com/browse/SNOW-258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,93 @@
+USE SCHEMA {{database_name}}.synapse_aggregate; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE DYNAMIC TABLE UNIQUE_USER_UPLOADS
+    (
+	    GRANULARITY VARCHAR(16777216) COMMENT 'The dimension of the aggregate (e.g., YEARLY, MONTHLY, DAILY).',
+	    UNIQUE_USER_COUNT INT COMMENT 'The number of distinct unique users uploading during the aggregation period.',
+	    YEAR NUMBER(38,0) COMMENT 'Year of the aggregation period.',
```
thomasyu888, line 7: Bringing the discussion here so Phil can chime in on 

* aggregate_year
* event_year
* rollup_year

as the choice here

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,93 @@
+USE SCHEMA {{database_name}}.synapse_aggregate; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE DYNAMIC TABLE UNIQUE_USER_UPLOADS
```
thomasyu888, line 3: Similar to the `unque_user_count`, maybe this is named `uploads...`

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,93 @@
...
+        SELECT
+            YEAR(record_date)  AS year,
+            MONTH(record_date) AS month,
+            DAY(record_date)   AS day,
+            COUNT(DISTINCT user_id) AS unique_user_count,
+            GROUPING(day)   AS g_day,
+            GROUPING(month) AS g_month,
+            GROUPING(year)  AS g_year
+        FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILEUPLOAD
+        WHERE RECORD_DATE <= MAX(RECORD_DATE) - 1
```
thomasyu888, line 30: I think we can remove the `-1`

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,93 @@
...
+	    YEAR NUMBER(38,0) COMMENT 'Year of the aggregation period.',
+        MONTH NUMBER(38,0) COMMENT 'Month of the aggregation period.',
+        DAY NUMBER(38,0) COMMENT 'Day of the aggregation period.',
+        AGGREGATE_PERIOD_START DATE COMMENT 'The time when any change to the team was made (e.g. update of the team or a change to its members).',
+        AGGREGATE_PERIOD_STOP DATE COMMENT 'The time when any change to the team was made (e.g. update of the team or a change to its members).',
+	    SNAPSHOT_DATE DATE COMMENT 'Date the aggregation was calculated and stored in the table.',
+        IS_COMPLETE BOOLEAN COMMENT 'If true, then the aggregation period is complete.'
+    )
+    TARGET_LAG = '1 day'
+    WAREHOUSE = compute_xsmall
```
thomasyu888, line 16: Do we want to look into DYNAMIC table update strategy.  Right now i think this table gets re-written to, but really there's no need to re-compute aggregates that have already been done.  For example, 2024 aggregates only need to be computed once, we don't need to be running the aggregates for those.

Does dynamic tables do some sort of upsertting where it will only include rows that are unchanged (minus snapshot_date)

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
philerooski: Let's begin the naming nit megathread 😄 

Since we'll be reusing many of these columns across our aggregate tables this is actually pretty important.

**day / month / year** – The idea here is to avoid SQL keywords? I didn't have any issues doing a query like this:
```
select day, month, year, day(aggregate_period_start)
from unique_user_uploads;
```
so I wonder how important it is that we change this. Analysts will be typing these column names a lot so it would be nice to keep them dead simple. One idea is to simply add an underscore, either as a prefix or suffix, e.g. `day_`, `month_`, `year_`. Another idea is to make them more precise: `day_of_month`, `month_of_year`, `year...` although the pattern sort of falls apart once we get to "year". Suggestions for year? Otherwise I'm in favor of `aggregate_year`. Perhaps we can reserve the `aggregate_` prefix for columns which are related to the time window of the aggregate. 

**granularity** – These values (daily, monthly, yearly) are all periods, so maybe we should rename this to `period` or `aggregate_period`. That would gel nicely with related columns `aggregate_period_start` and `aggregate_period_stop`.

**aggregate_period_start / aggregate_period_stop** – I realize I talked about using the `aggregate_` prefix as a way of identifying columns related to the aggregate time window, but if we aren't doing that then I think "aggregate" is implied by this being an aggregate/counts table, and we could simplify this as `period_start` and `period_stop`. Super nit: since we're talking about a range of dates, it might be more accurate to talk about the the range "start" and "end" or "begin" and "end". Something like `period_start` and `period_end`.

**snapshot_date** – How directly does this map back to "snapshots" as platform uses the term? If this is the date that a row was last updated, it feels quite different from an event snapshot. I don't want to repurpose "snapshot" in the column name unless we're talking about the same process.

**is_complete** – I think this could be more precise, particularly since we have data types like certified quiz where this could potentially have multiple interpretations, for example, if a user were able to save their quiz progress. `aggregate_is_complete`, `period_is_complete`, `aggregate_period_is_complete` are a few possibilities out of many.

**unique_user_count** – I saw in the presenter notes that you didn't want to use `unique_`. How come? We also might want to make this more precise if we modify the table name:

**UNIQUE_USER_UPLOADS** More generally, this is an aggregate table for each user. We eventually may want to add additional aggregates like downloads, submissions, etc. So a more general-purpose name would be `USER`, as I describe in the design doc [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3998548038/SNOW-187+Gold-layer+table+framework+aggregates#Tables). This might not be so important to change in this PR. We can always deprecate this table if we want all user-level aggregates in a single table, although deprecation comes with its own headaches.

thomasyu888: > day / month / year – The idea here is to avoid SQL keywords? I didn't have any issues doing a query like this:

If you look online, it's bad practice to use reserved keywords as column headers as that can lead to ambiguity, I'm ok with `aggregate_`

> quarter column
Yes, that would be good to add as a rollup

> `period_start` / `period_stop`
I like this

> `snapshot_date`
Can you propose another name?  I had the same thought that `snapshot`ing means something specific BUT also currently, in Synapse, there are two events that drive is so....

> `is_complete` – I think this could be more precise...
I like this, I'll defer to you guys to come up with the name here.

> unique_user_count – I saw in the presenter notes that you didn't want to use unique_. How come? We also might want to make this more precise if we modify the table name

`unique` is a SQL reserved keyword, so Kevin, as the user of the table, wanted something that didn't start with SQL reserved key words.

> UNIQUE_USER_UPLOADS More generally, this is an aggregate table for each user. 

Hmm, I think this may require more discussion.

philerooski: > snapshot_date

>> Can you propose another name?

`snapshot_date` is defined as `CURRENT_DATE` here, and since I think it's unavoidable to rebuild the table from scratch each time, this makes this column redundant with the date of the last table refresh. I recommend removing it for now.

> unique is a SQL reserved keyword, so Kevin, as the user of the table, wanted something that didn't start with SQL reserved key words.

In the specific case of a "user count", I think it's implied that each user is unique. So merely "user_count" or "users" would work here. Although we would definitely want to put the word "upload" somewhere in the column name if we were to make this table more generalizable than a single metric.

But there are scenarios where it would be more descriptive to specify "unique" or "distinct". For example, if we were computing aggregates for each Synapse entity and wanted to know how many versions of this entity exist which have a distinct md5. That would be different from merely counting the entity versions, so it would make sense to name the column something like `unique_versions` (`unique_version_count`) rather than just `versions` (`version_count`).

> UNIQUE_USER_UPLOADS More generally, this is an aggregate table for each user.

>> Hmm, I think this may require more discussion...

I was wrong about this. This isn't an aggregate table for each user. It's just an aggregate table across all of Synapse.

thomasyu888: > it's unavoidable to rebuild the table from scratch each time, this makes this column redundant with the date of the last table refresh. I recommend removing it for now.

Thanks @philerooski can we dive into this?  For example

```
GRANULARITY, USER_COUNT, START, STOP, IS_COMPLETE
DAILY,4,05/04/2025, 05/04/2025, TRUE
DAILY,5,05/05/2025, 05/04/2025, TRUE
MONTHLY,5,05/05/2025, 05/30/2025, FALSE
YEARLY,30,01/01/2025, 12/31/2025, FALSE
```

Tomorrow it'll be

```
GRANULARITY, USER_COUNT, START, STOP, IS_COMPLETE
DAILY,4,05/04/2025, 05/04/2025, TRUE  <- This doesn't change
DAILY,5,05/05/2025, 05/05/2025, TRUE  <- This doesn't change
DAILY,5,05/06/2025, 05/06/2025, TRUE <- this is appended
MONTHLY,7,05/01/2025, 05/30/2025, FALSE <- This is updated
YEARLY,34,01/01/2025, 12/31/2025, FALSE <- this is updated
```

> In the specific case of a "user count", I think it's implied that each user is unique. So merely "user_count" 

I like that.

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,93 @@
...
+
+CREATE OR REPLACE DYNAMIC TABLE UNIQUE_USER_UPLOADS
+    (
+	    GRANULARITY VARCHAR(16777216) COMMENT 'The dimension of the aggregate (e.g., YEARLY, MONTHLY, DAILY).',
+	    UNIQUE_USER_COUNT INT COMMENT 'The number of distinct unique users uploading during the aggregation period.',
+	    YEAR NUMBER(38,0) COMMENT 'Year of the aggregation period.',
+        MONTH NUMBER(38,0) COMMENT 'Month of the aggregation period.',
+        DAY NUMBER(38,0) COMMENT 'Day of the aggregation period.',
+        AGGREGATE_PERIOD_START DATE COMMENT 'The time when any change to the team was made (e.g. update of the team or a change to its members).',
+        AGGREGATE_PERIOD_STOP DATE COMMENT 'The time when any change to the team was made (e.g. update of the team or a change to its members).',
```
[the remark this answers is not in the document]
philerooski: We ought to be able to use an inline constraint. e.g.,
```
...
YEAR NUMBER(38,0)
  PRIMARY KEY
  COMMENT 'Year of the aggregation period.',
...
```

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
philerooski: I don't think the dynamic table is smart enough to avoid recomputing the aggregates. See the "optimizing aggregates" section in my design doc [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3998548038/SNOW-187+Gold-layer+table+framework+aggregates#Example%3A-O3-KR2) for how we can cache an intermediate computation so that we can build the aggregate table faster. (This is overkill for now, imo).

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.50.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,94 @@
...
+            YEAR(record_date)  AS year,
+            MONTH(record_date) AS month,
+            DAY(record_date)   AS day,
+            COUNT(DISTINCT user_id) AS unique_user_count,
+            GROUPING(day)   AS g_day,
+            GROUPING(month) AS g_month,
+            GROUPING(year)  AS g_year
+        FROM {{database_name}}.SYNAPSE.FILEUPLOAD --noqa: JJ01,PRS,TMP
+        -- exclude today’s partial data to avoid confusion in the final table
+        WHERE RECORD_DATE < (SELECT MAX(RECORD_DATE) FROM {{database_name}}.SYNAPSE.FILEUPLOAD) --noqa: JJ01,PRS,TMP
```
jaymedina, line 31: Example of what the latest record would be for the query run on May 12, 2025:
image

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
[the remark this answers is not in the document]
jaymedina: It doesn't look like we can use inline or out-of-line constraints for dynamic tables :< 

I've updated the column comments to mention which of the columns belong to the composite PK, though. I read that PKs don't actually enforce unique-ness on Snowflake so this isn't a huge loss.

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
jaymedina: To finalize some things:

* granularity -> period_type (< this OK @philerooski ? "period" sounds like I should expect a date object)
* aggregate_period_start -> period_start
* aggregate_period_stop -> period_end 
* is_complete -> period_is_complete
* unique_user_count -> user_count

Are we all good with these? Also, is there an issue with having `unique` in the table name, too?

After we finalize the column names, I'll update this PR with a quarterly rollup.

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
jaymedina: `user_uploads`?

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
philerooski: @jaymedina What do you think of prefixing these meta-type columns with "aggregate" or "agg" (since "aggregate" is already implied by the schema name and it's more pleasant to type) to differentiate them from our metric-type columns like `user_count`? 

e.g.,

* agg_day
* agg_month
* agg_year
* agg_period
* agg_period_start
* agg_period_end
* agg_period_is_complete

The idea being that if a user is looking to filter down their results in a WHERE clause they can type "agg" in snowsight, hit tab, and have all possibilities suggested at once. It could be cleaner than a mixed naming scheme like:

* agg_day
* agg_month
* agg_year
* period_type
* period_start
* period_end
* period_is_complete

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
philerooski: I think if we're not ready to commit to "this is our aggregate table where all Synapse-wide metrics live" then `USER_UPLOADS` makes sense.

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
jaymedina: agg_day
    agg_month
    agg_year
    agg_period
    agg_period_start
    agg_period_end
    agg_period_is_complete

Works for me!

> if a user is looking to filter down their results in a WHERE clause they can type "agg" in snowsight, hit tab, and have all possibilities suggested at once

Good mode of thinking.

I'll update now!

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
philerooski: Whoops, I commented before I saw the previous comment. Let me respond...

jaymedina: > we're not ready to commit to "this is our aggregate table where all Synapse-wide metrics live" 

We're not? Can I get more context on this? This is our aggregate table where Synapse-wide user uploads are totaled across time.

philerooski: My primary take is that we don't yet know how to best organize our aggregate data so that it optimizes both developer efficiency (to add new aggregate metric columns) and analyst efficiency (to easily find and query aggregate metric data), so the name of this table is not so important at this point. Name it whatever you want.

My secondary take is that I think this metric (`user_count`) is likely to be absorbed into a more holistic aggregate table: like a table which contains multiple aggregates derived from file upload data, or a table which contains aggregates across Synapse more generally. Aggregate tables like `{context}_{metric}_{dimension}` is probably too specific, and could result in a brand new table for just one or two aggregate metrics. This would hurt analyst efficiency.

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
jaymedina: That's correct.

Because we're using `GROUP BY ROLLUP`, our aggregate dynamic tables can only be refreshed in full. Which means we need to re-compute all the aggregates each time.

In the future it would be nice to have agg tables that capture rates of change, so what did the YEARLY agg for 2024 look like in Q1 vs Q4? This is different from comparing Q1 vs. Q4 QUARTERLY aggs.

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
jaymedina: > we don't yet know how to best organize our aggregate data so that it optimizes both developer efficiency (to add new aggregate metric columns) and analyst efficiency (to easily find and query aggregate metric data)

I see, that makes sense. In that case, I'll just rename this to `user_uploads`. I don't feel strongly about it right now.

Slightly off-topic but I'd like to give my take on this:

> Aggregate tables like {context}_{metric}_{dimension} is probably too specific, and could result in a brand new table for just one or two aggregate metrics. This would hurt analyst efficiency.

As far as future design, I think analyst efficiency directly ties with query efficiency, and query efficiency directly ties with how small of a scope the aggregate tables are. It might be more efficient to have many small-scoped tables that can be joined by foreign keys, than to have large aggregate tables of different contexts, but this is probably one of those things that we'll see a better picture of when we start writing POCs. No final opinion on this, just giving my 2 cents!

thomasyu888: @philerooski / @jaymedina , let's have a way to clearly indicate tables that are "experiment" vs not, but I appreciate this current topic of discussion!

jaymedina: @thomasyu888 TBH the entire warehouse is in experimental mode 😆 I don't want to deter users from working with these agg tables in their current state. Chances are their data structure will change, but the aggregate calculations in them are still valid.

thomasyu888: @jaymedina that's fair. We'll just have to be careful when we move tables around and/or deprecate tables that we clearly outline that and let users know to use the newer tables.

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.49.0__unique_user_uploads_aggregate.sql
jaymedina: I'll resolve this once I've finished implementing the quarterly rollup! 🔢

jaymedina:
@philerooski @thomasyu888 

The new quarterly dimension has been [added](https://github.com/Sage-Bionetworks/snowflake/pull/201/commits/42d502fa16b647ed1cfe9209f25ea5ea1fb45b68) and you can view the table on Snowsight.

Apparently Snowflake doesn't honor the ordering in the `ORDER BY` clause when introducing dynamic tables so I'm going to remove that from the ETL. I was hoping to see something like below, but you'll see that things are ordered differently when you go on SNowsight. Good to know for when we write DTs in the future, there's no point in using `ORDER BY` for the final query (still useful for the `*latest` ROW_NUMBER calculations, though).
image

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.50.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,104 @@
+USE SCHEMA {{database_name}}.synapse_aggregate; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE DYNAMIC TABLE USER_UPLOADS
+    (
+	    AGG_PERIOD VARCHAR(16777216) COMMENT 'The dimension of the aggregate (e.g., YEARLY, MONTHLY, DAILY). Part of composite PK.',
+	    USER_COUNT INT COMMENT 'The number of distinct unique users uploading during the aggregation period.',
+	    AGG_YEAR NUMBER(38,0) COMMENT 'Year of the aggregation period.',
```
BryanFauble, line 7: nit: Consistent tabbing with the next few lines

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.50.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,104 @@
...
+        AGG_QUARTER NUMBER(38,0) COMMENT 'Quarter of the aggregation period.',
+        AGG_MONTH NUMBER(38,0) COMMENT 'Month of the aggregation period.',
+        AGG_DAY NUMBER(38,0) COMMENT 'Day of the aggregation period.',
+        AGG_PERIOD_START DATE COMMENT 'The start date of the aggregation period. Part of composite PK.',
+        AGG_PERIOD_END DATE COMMENT 'The stop date of the aggregation period. Part of composite PK.',
+        AGG_PERIOD_IS_COMPLETE BOOLEAN COMMENT 'If true, then the aggregation period is complete.'
```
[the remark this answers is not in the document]
jaymedina: Yep, the ER diagram was made as an initial draft of the solution. Since then, review comments have been made with minor suggestions to the data structure. I'll note that in the description of the PR.

philerooski (commented):
Left some last-minute comments.

Overall this is looking great. The `agg_period*` columns are a real nice touch that make filtering for specific aggregates significantly easier than relying on the columns provided by GROUP BY ROLLUP.

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.50.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,104 @@
...
+    -- Step 3: Compose the final table by ordering the columns for easier reading and adding the completion column
+    SELECT
+        agg_period,
+        user_count,
+        agg_year,
+        agg_quarter,
+        agg_month,
+        agg_day,
+        agg_period_start,
+        agg_period_end,
```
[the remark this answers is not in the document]
jaymedina: Yes 👍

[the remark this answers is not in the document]
jaymedina: Actually since `user_count` is what the table is centered around, what do you think of this order:

```
user_count,
        agg_period,
        agg_year,
        agg_quarter,
        agg_month,
        agg_day,
        agg_period_start,
        agg_period_end,
```

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.50.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,104 @@
...
+                WHEN g_year = 0 AND g_quarter = 0 AND g_month = 0 AND g_day = 1 THEN 'MONTHLY'
+                WHEN g_day = 0 THEN 'DAILY'
+            END AS agg_period,
+
+            -- 3. Create `agg_period_start` column
+            CASE
+                -- First day of the year
+                WHEN g_year = 0 AND g_quarter = 1 AND g_month = 1 AND g_day = 1 THEN DATE_FROM_PARTS(agg_year, 1, 1)
+                -- First day of the quarter
+                WHEN g_year = 0 AND g_quarter = 0 AND g_month = 1 AND g_day = 1 THEN DATE_FROM_PARTS(agg_year, (agg_quarter * 3) - 2, 1)
```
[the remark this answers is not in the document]
jaymedina: I ran the query with the QUARTER function without modifying these and don't see a change in results:

image

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.50.0__unique_user_uploads_aggregate.sql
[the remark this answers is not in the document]
philerooski: That's fine. I'm used to seeing the metadata/context columns take precedence, e.g., like when we create a file view in Synapse. A [cursory search of SO threads](https://stackoverflow.com/questions/894522/is-there-any-reason-to-worry-about-the-column-order-in-a-table?rq=3) suggests that it's generally a good idea to have your primary key columns first, but I'm not aware of any standard for column ordering.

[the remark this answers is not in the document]
jaymedina: > I'm used to seeing the metadata/context columns take precedence, e.g., like when we create a file view in Synapse

OK! I will stick with the original suggestion

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.50.0__unique_user_uploads_aggregate.sql
philerooski: probably a tab vs. spaces thing 😄 If you delete the indent for all these lines and re-do the indent, that ought to fix it

synapse_data_warehouse/synapse_aggregate/dynamic_tables/V2.50.0__unique_user_uploads_aggregate.sql
```diff
@@ -0,0 +1,104 @@
...
+        agg_month,
+        agg_day,
+        agg_period_start,
+        agg_period_end,
+    
+        -- 5. Mark the period as complete once today's date is past the stop
+        (CURRENT_DATE > agg_period_end) AS agg_period_is_complete
+
+    FROM agg_period_calculations
+    WHERE agg_year is NOT NULL; -- drop the all-NULL “grand total” row
```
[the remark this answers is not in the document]
jaymedina: > It could be valuable to include how many users have uploaded data since 2022

Fair point! What do you think of this implementation? A new 'grain' called `ALL TIME` that will capture the cumulative total of unique user uploads since the earliest record date.

image

philerooski (approved):

PR #202, opened 2025-05-06
[SNOW-269] SYNAPSE_EVENT schema

philerooski:
# Problem

Ticket: [SNOW-269](https://sagebionetworks.jira.com/browse/SNOW-269)

# Solution

1. Decide whether we want to create synapse_all_events (or TBD name) schema
    * I create a SYNAPSE_EVENT schema (following the singular convention in SQL object names) and roles/grants for both tables and dynamic tables.
2. Create necessary documentation
    * I wrote documentation which explains how to set up roles and grants for any general object type. You can read that [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4115366084/Managing+Object+Privileges).

# Testing

We can validate the existence of our dummy table and the relationships between our database roles, but account-level grants (such as future grants and grants to account roles) and grants on anything that is not a database role are not run until we merge into main.
```
use role data_engineer;

-- verify that we have database roles for admin, developer, analyst, and table/read
show database roles in database synapse_data_warehouse_dev_snow_269_synapse_event_schema;
SELECT * FROM TABLE(RESULT_SCAN(LAST_QUERY_ID())) WHERE "name" like 'SYNAPSE_EVENT%';

-- verify that admin db role has ownership of all database roles
show grants to database role synapse_data_warehouse_dev_snow_269_synapse_event_schema.synapse_event_all_admin;
-- verify that the proxy admin has ownership and usage of admin db role
show grants on database role synapse_data_warehouse_dev_snow_269_synapse_event_schema.synapse_event_all_admin;

-- verify that the developer db role has usage of the analyst db role
show grants to database role synapse_data_warehouse_dev_snow_269_synapse_event_schema.synapse_event_all_developer;

-- verify that the analyst db role has usage of the table/read db role
show grants to database role synapse_data_warehouse_dev_snow_269_synapse_event_schema.synapse_event_all_analyst;

-- verify that our dummy table has values
select * from synapse_data_warehouse_dev_snow_269_synapse_event_schema.synapse_event.dummy_table;
```

 

[SNOW-269]: https://sagebionetworks.jira.com/browse/SNOW-269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


synapse_data_warehouse/synapse_event/V2.49.0__create_synapse_event_schema.sql
```diff
@@ -0,0 +1,3 @@
+USE DATABASE {{ database_name }}; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE SCHEMA SYNAPSE_EVENT;
```
thomasyu888, line 3: Can we add a detailed description of what tables will exist in this schema?

philerooski: Sure, I can add it to either Confluence documentation or the Jira ticket. I don't think burying high-level outlines in the source code is very useful.

thomasyu888: No I mean you can use COMMENT and if should show up on snowflake UI

thomasyu888 (commented):
🔥 LGTM!  I'll defer to @jaymedina for final review

admin/ownership_grants/V1.15.1__synapse_event_ownership.sql
```diff
@@ -0,0 +1,22 @@
+-- SYNAPSE_DATA_WAREHOUSE
+GRANT OWNERSHIP
+    ON SCHEMA SYNAPSE_DATA_WAREHOUSE.SYNAPSE_EVENT
+    TO DATABASE ROLE SYNAPSE_DATA_WAREHOUSE.SYNAPSE_EVENT_ALL_ADMIN
+    REVOKE CURRENT GRANTS;
```
jaymedina, line 5: What is the reason for using `REVOKE CURRENT GRANTS` here? Just wondering because I didn't see it when `SYNAPSE_AGGREGATE` was introduced

image

Also, we are missing USAGE/MONITOR grants to `*ALL_DEVELOPER`

synapse_data_warehouse/synapse_event/V2.49.0__create_synapse_event_schema.sql
```diff
@@ -0,0 +1,4 @@
+USE DATABASE {{ database_name }}; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE SCHEMA SYNAPSE_EVENT
+    COMMENT = 'A schema for Synapse event data. Event data is derived from raw data by retaining only the most recent snapshot for each index.';
```
jaymedina, line 4: > _Event data is derived from raw data by retaining only the most recent snapshot for each index._

No, these tables are capturing _all_ events, not just the most recent ones. They are historical tables that capture all changes, including deletions. We are just removing duplicate snapshots of the same event.

At least that is my understanding. cc @thomasyu888 and @danlu1 to confirm.

jaymedina (changes requested):
This is looking good and clean ✨ Thanks @philerooski! I'm doing an initial pass through and will revisit later today or tomorrow for a final review.

synapse_data_warehouse/synapse_event/V2.49.0__create_synapse_event_schema.sql
thomasyu888: That's right, these capture all historical events. It's confusing because the weekly snapshotting actually DOES contain all historical events for tables like node. 

If the snapshot_date wasn't updated by two events, it wouldn't be as complicated.

thomasyu888: I think the definition of "index" isn't defined here. But each table may have a different index that changes what the most recent snapshot is.

philerooski: @jaymedina 
> No, these tables are capturing all events, not just the most recent ones.

Are you confusing events with snapshots? The comment says we retain only the most recent snapshot for each index, which means there will be one snapshot retained for each event.

@thomasyu888 
> I think the definition of "index" isn't defined here. But each table may have a different index that changes what the most recent snapshot is.

I want to say that for each unique "event", which should be determined by some unique subset of columns (depending on the table), we keep the most recent snapshot. But I think "index" may be confusing here.

It might be less ambiguous to instead say:

> Event data is derived from raw data by retaining only the most recent snapshot for each distinct event identifier. The identifier of an event is determined by a subset of columns which uniquely identifies the occurrence or instance of an event.

jaymedina: > Are you confusing events with snapshots? 

No, I understand you can have multiple snapshots for a single event, which is what we're de-duplicating. I see now that you're referring to the event when you say _index_, rather than the node itself.

> Event data is derived from raw data by retaining only the most recent snapshot for each distinct event identifier. The identifier of an event is determined by a subset of columns which uniquely identifies the occurrence or instance of an event.

This rewrite is more clear to me, thanks!

thomasyu888 (approved):
🔥 LGTM!

jaymedina (approved):
Just did a final pass through and all changes lgtm! ✔️

PR #205, opened 2025-05-14
[SNOW-279] Add node_events table

philerooski:
# Problem

Ticket: [SNOW-279](https://sagebionetworks.jira.com/browse/SNOW-279)

# Solution

1. Create a table that can answer all the questions above without "duplicates" ✅ 
    * In contrast with the node_latest table, I take both the `id` and `version_number` as primary keys. This gives us the latest snapshot for each version of a Synapse entity. 
3. create a mermaid ERD diagram ❌ 
4. add primary key and foreign key constraints to snowflake ✅  
    * These are not supported in dynamic tables, but I have indicated the primary keys in the column comments.

# Testing

Testing via draft PR.


synapse_data_warehouse/synapse_event/V2.51.0__node_event_table.sql
```diff
@@ -0,0 +1,92 @@
...
+                DEFINING_SQL,
+                INTERNAL_ANNOTATIONS,
+                VERSION_HISTORY,
+                PROJECT_STORAGE_USAGE
+            FROM
+                {{ database_name }}.synapse_raw.nodesnapshots --noqa: TMP
+            WHERE
+                snapshot_timestamp >= CURRENT_TIMESTAMP - INTERVAL '30 DAYS'
+            QUALIFY ROW_NUMBER() OVER (
+                    PARTITION BY id, version_number
```
thomasyu888, line 50: I know this is draft mode, but take a look at all the questions we would like to answer.

Specifically, entities can be updated without bumping the version - I think there is a question surrounding that

jaymedina: We also want to capture DELETE events, right? Something I noticed is that there are no DELETE events with `version_number` as part of the entity metadata. 

image

I looked throuhg some of these and they actually don't have much useful information, not even `node_type` 🤔 Might be something worth asking the platform team about too.

image

thomasyu888: I would try and answer all the questions as listed in the Jira ticket with this table.

philerooski:
@jaymedina Moving this to its own thread:

> We also want to capture DELETE events, right?

We do capture DELETE events. `version_number` = `null` is considered its own partition, so we get the latest snapshot of the delete event included in the table.

This is how I verified:
```
-- do we capture all DELETE events from the snapshot table (they don't have a `version_number`)
select * from node_event where change_type = 'DELETE';
-- 2977192
select
  count(distinct id)
from
  synapse_data_warehouse.synapse_raw.nodesnapshots
where
  change_type = 'DELETE' and snapshot_date > current_date - INTERVAL '30 days';
-- 2977168, difference is from lag
select
  count(*)
from
  node_event
where
  change_type = 'DELETE' and snapshot_date > current_date - INTERVAL '30 days';
```

synapse_data_warehouse/synapse_event/dynamic_tables/V2.52.0__node_event_table.sql
```diff
@@ -0,0 +1,92 @@
...
+                DEFINING_SQL,
+                INTERNAL_ANNOTATIONS,
+                VERSION_HISTORY,
+                PROJECT_STORAGE_USAGE
+            FROM
+                {{ database_name }}.synapse_raw.nodesnapshots --noqa: TMP
+            WHERE
+                snapshot_timestamp >= CURRENT_TIMESTAMP - INTERVAL '30 DAYS'
+            QUALIFY ROW_NUMBER() OVER (
+                    PARTITION BY id, version_number, change_type
```
thomasyu888, line 50: You are able to "update" the same version over and over.  Maybe look at etag?

synapse_data_warehouse/synapse_event/V2.51.0__node_event_table.sql
philerooski: @thomasyu888

> Specifically, entities can be updated without bumping the version - I think there is a question surrounding that

`version_number` uniquely enumerates updates to a node, even when a Synapse user is updating a previous version of the entity. See the Slack thread here: https://sagebionetworks.slack.com/archives/C05PBT2FUHF/p1747334513248759

synapse_data_warehouse/synapse_event/dynamic_tables/V2.52.0__node_event_table.sql
philerooski: I think this answers your concerns? https://github.com/Sage-Bionetworks/snowflake/pull/205#discussion_r2093382761

philerooski:
@jaymedina @thomasyu888 

This is ready for review.

A good starting place (after reviewing the code) is to look through the "Testing" section in my PR description.

There's still an open question about how we should handle CREATE events; specifically, some nodes have multiple CREATE events with different `modified_on` dates. This means that each of these nodesnapshot records is considered an event (`id`/`version_number`/`change_type`/`modified_on` are unique). Although, fortunately, because the `created_on` timestamp doesn't seem to change across these "duplicate" events, we still have the original time the node was created. See the exploratory work I did for Q1 (look for comment "We might expect just a single CREATE event for any node...") for more details.

If these "duplicates" are an artifact of the data warehouse, then we can qualify CREATE events differently from UPDATE and DELETE events and only keep one CREATE event per node. But if these "duplicates" actually contain valuable information or are an upstream bug then we can either keep the SQL as-is or have Platform fix the issue.

thomasyu888 (commented):
🔥 Thanks for all that exploration work and answering those questions.  

I had a some questions on the exploratory work and I'll defer to @jaymedina for final review. 
1. Why did you choose to use `change_timestamp` instead of `created_on`?
2. For the total file size, do you think we need to do some massaging of the `filesnapshots` table to get `fileevents`?  Overall, there was a schema change in Synapse some time in 2019, so that's why certain versions aren't backfilled

philerooski:
@thomasyu888 
> Why did you choose to use change_timestamp instead of created_on?

What's the context for this question? I don't use `change_timestamp` in the table definition.

> For the total file size, do you think we need to do some massaging of the filesnapshots table to get fileevents? Overall, there was a schema change in Synapse some time in 2019, so that's why certain versions aren't backfilled

Definitely. I'd like everything in my Q4 exploratory work to be better understood before we start working with the filesnapshots data. I wouldn't be surprised if there are some other oddities I could dig up if I spent more time exploring the filesnapshots table.

synapse_data_warehouse/synapse_event/dynamic_tables/V2.53.0__node_event_table.sql
```diff
@@ -0,0 +1,100 @@
...
+                PROJECT_STORAGE_USAGE
+            FROM
+                {{ database_name }}.synapse_raw.nodesnapshots --noqa: TMP
+            WHERE
+                -- only `id`, `change_*`, and `snapshot_*` columns are defined regardless of `change_type`
+                id is not null
+                and change_type is not null
+            QUALIFY ROW_NUMBER() OVER (
+                    -- `id` and `version_number` provide at least one record for each Synapse entity version.
+                    -- `change_type` ensures that simultaneous CREATE+UPDATE events are both retained.
```
jaymedina, line 53: nit: `CREATE+UPDATE+DELETE` events are all retained

synapse_data_warehouse/synapse_event/dynamic_tables/V2.53.1__node_event_table_comments.sql
```diff
@@ -0,0 +1,38 @@
...
+COMMENT ON COLUMN node_event.change_timestamp IS 'The time when the change (created/updated/deleted) on the node is pushed to the queue for snapshotting.';
+COMMENT ON COLUMN node_event.change_user_id IS 'The unique identifier of the user who made the change to the node.';
+COMMENT ON COLUMN node_event.node_type IS 'The type of the node. Allowed node types are : project, folder, file, table, link, entityview, dockerrepo, submissionview, dataset, datasetcollection, materializedview, virtualtable.';
+COMMENT ON COLUMN node_event.parent_id IS 'The unique identifier of the parent in the node hierarchy.';
+COMMENT ON COLUMN node_event.benefactor_id IS 'The identifier of the (ancestor) node which provides the permissions that apply to this node. Can be the id of the node itself.';
+COMMENT ON COLUMN node_event.project_id IS 'The project where the node resides. It will be empty for the change type DELETE.';
+COMMENT ON COLUMN node_event.created_on IS 'The creation time of the node.';
+COMMENT ON COLUMN node_event.created_by IS 'The unique identifier of the user who created the node.';
+COMMENT ON COLUMN node_event.modified_on IS 'The most recent change time of the node.';
```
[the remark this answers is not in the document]
philerooski: Actually I just forgot to update the comments. I think I understand the reasoning behind primary keys not being null (in general), and some SQL dialects (not Snowflake) enforce this explicitly. Seems like the typical workaround is to generate a unique hash from the composite primary key columns (the hash is never null) and use that as the primary key. Even if we did that to do things by the book, we would still need to communicate somehow that these 4 columns uniquely identify the record in the table.

I like the suggestion by GPT:
> In Snowflake, primary‐key (and unique) constraints are purely metadata—they’re not enforced by the engine—so you can declare a composite key over columns that may be nullable, and rely on documentation and downstream tools to honor it.

1. Declare the composite PK (metadata only). Even if `version_number` and `modified_on` can be NULL, you can still add the constraint. [I tested this in Snowflake and yes, it is possible to add primary key constraints to a dynamic table].
2. Use COMMENT ON to bake your data‐model intent right into the table DDL:
```
-- Table‐level comment
COMMENT ON TABLE node_event
  IS 'Each row is uniquely identified by (id, version_number, change_type, modified_on), even if some of these may be NULL.';

-- Column‐level comments
COMMENT ON COLUMN node_event.id
  IS 'Primary key (composite). ...';


COMMENT ON COLUMN node_.version_number
  IS 'Primary key (composite). ...may be NULL if `change_type` = \'DELETE\'.';
```
3. (Optional). Add a hash column and declare it unique.
4. Surface the composite primary keys in other documentation (e.g., Confluence)

thomasyu888:
> What's the context for this question? I don't use `change_timestamp` in the table definition.

@philerooski, I was referring to all the SQL queries you wrote to answer the questions.

philerooski:
@jaymedina 

> Should we talk to platform ab why DELETE events for nodes in nodesnapshots aren't captured with any metadata?

Yeah, definitely! The DELETE events are mostly useless and sometimes contradictory. Check out this example:

```
-- The first DELETE snapshot was taken at 2025-05-20 02:30:21.755
-- But this is followed by an (apparently) duplicate snapshot of the
-- initial CREATE event, followed by a later DELETE event at 2025-05-20 14:48:28.180
select
  *
from
  synapse_data_warehouse.synapse_raw.nodesnapshots
where
  id = 66612361
  and snapshot_date > '2025-05-19' -- cheating a little here, since we know this was created on 2025-05-20
order by
  snapshot_timestamp;
```

How are we supposed to know when this entity was deleted? The `change_timestamp` is useless (it's nearly identical with the `created_on` time) and the only other timestamp information we have about this event is the `snapshot_timestamp`.

Also see this related ticket (reported by you): https://sagebionetworks.jira.com/browse/PLFM-8747

A more general confusion I think is that sometimes we get change events, but no change in the `change_timestamp`. We saw this with DELETE events in the examples above, but I'm pretty sure I've seen this with UPDATE events as well. I wonder if it's related to the UPDATE event happening to a parent, rather than the entity itself? Just a wild guess -- we need to explore this further.

philerooski:
@thomasyu888 

> >What's the context for this question? I don't use change_timestamp in the table definition.

> @philerooski, I was referring to all the SQL queries you wrote to answer the questions.

I should have used `created_on` in Q3. I substituted it in my worksheet and still got the same number across node_event / nodesnapshots. I didn't see `change_timestamp` being used to answer any of the other questions.

I also used `change_timestamp` in some of my miscellaneous queries, but those are more exploratory and not as relevant to verifying the correctness of the table definition.

synapse_data_warehouse/synapse_event/dynamic_tables/V2.53.1__node_event_table_comments.sql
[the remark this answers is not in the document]
jaymedina: I'm resolving this because of what you showed me, that PK constraints _can_ be added to dynamic tables! Great find!

synapse_data_warehouse/synapse_event/dynamic_tables/V2.53.0__node_event_table.sql
```diff
@@ -0,0 +1,65 @@
...
+                -- only `id`, `change_*`, and `snapshot_*` columns are defined regardless of `change_type`
+                id is not null
+                and change_type is not null
+            QUALIFY ROW_NUMBER() OVER (
+                    -- `id` and `version_number` provide at least one record for each Synapse entity version.
+                    -- `change_type` ensures that simultaneous events are retained, regardless of their `change_type`.
+                    -- `modified_on` enables us to capture updates to Synapse Table entities and other
+                    -- update scenarios where the `version_number` does not change (like when we update annotations
+                    -- without incrementing the version). 
+                    PARTITION BY id, version_number, change_type, modified_on
```
BryanFauble, line 57: Should we use `etag` within the `PARTITION BY` statement rather than `modified_on`? My thought is based on the wording below that each etag update indicates that "Something" changed:

> Synapse employs an Optimistic Concurrency Control (OCC) scheme to handle concurrent updates. Since the E-Tag changes every time an entity is updated it is used to detect when a client's current representation of an entity is out-of-date.

BryanFauble (approved):
Changes LGTM!

synapse_data_warehouse/synapse_event/dynamic_tables/V2.53.0__node_event_table.sql
philerooski: @BryanFauble

I think they're equivalent: https://sagebionetworks.slack.com/archives/C05PBT2FUHF/p1747426131028369?thread_ts=1747334513.248759&cid=C05PBT2FUHF

thomasyu888 (approved):
🔥 LGTM! great work here

synapse_data_warehouse/synapse_event/V2.51.0__node_event_table.sql
philerooski: Addressed by [d9fc4b6](https://github.com/Sage-Bionetworks/snowflake/pull/205/commits/d9fc4b677cccbaf0c023693a20a451e8076aac94).

synapse_data_warehouse/synapse_event/dynamic_tables/V2.52.0__node_event_table.sql
philerooski: Addressed by [d9fc4b6](https://github.com/Sage-Bionetworks/snowflake/pull/205/commits/d9fc4b677cccbaf0c023693a20a451e8076aac94).

philerooski:
I filed a ticket documenting the data quality issue around potentially erroneous CREATE events: https://sagebionetworks.jira.com/browse/PLFM-9049

Fortunately, this seems to affect just ~0.2% of nodes, although it does detract from the integrity of CREATE events as a whole: Is this _actually_ a CREATE event? Or just an artifact of the data warehouse?

For example, determining the number of nodes created during a given time period can expect ~0.2% margin of error (most of these cases have just 2 CREATE events).

Here's how I determined how many nodes are affected:
```
-- Cases where we have multiple CREATE events, but
--   * no DELETE events
--   * no instances where this node was trashed (benefactor_id = '1681355')
--   * at least one UPDATE event
--   * at least one CREATE happened after at least one UPDATE
--   * not all CREATE events have the same `modified_on` value
-- ~83k rows (0.2% of all nodes)
WITH 
  -- 1) find all IDs that had a DELETE (node_type will be NULL for those rows)
  delete_ids AS (
    SELECT DISTINCT id
    FROM node_event
    WHERE change_type = 'DELETE'
  ),
  -- 2) pull only events for IDs that never had a DELETE
  filtered_events AS (
    SELECT *
    FROM node_event ne
    WHERE ne.id NOT IN (SELECT id FROM delete_ids)
  )
SELECT
  id,
  node_type,
  -- count of CREATEs for the HAVING and ordering
  SUM(CASE WHEN change_type = 'CREATE' THEN 1 ELSE 0 END) AS n_creates
FROM
  filtered_events
GROUP BY
  id,
  node_type
HAVING
  -- at least two CREATEs
  SUM(CASE WHEN change_type = 'CREATE' THEN 1 ELSE 0 END) > 1
  -- at least one UPDATE
  AND SUM(CASE WHEN change_type = 'UPDATE' THEN 1 ELSE 0 END) > 0
  -- no unwanted benefactor
  AND SUM(CASE WHEN benefactor_id = '1681355' THEN 1 ELSE 0 END) = 0
  -- CREATE events have more than one distinct modified_on
  AND COUNT(DISTINCT CASE WHEN change_type = 'CREATE' THEN modified_on END) > 1
  -- at least one CREATE happened after at least one UPDATE
  AND MAX(CASE WHEN change_type = 'CREATE' THEN modified_on END)
      > MIN(CASE WHEN change_type = 'UPDATE' THEN modified_on END)
ORDER BY
  n_creates DESC;
```

Here's a few more examples of this phenomenon:
```
select
  *
from
  node_event
where
  id = 4489
  --id = 61982402
  --id = 12216422
  --id = 11923688
order by
  snapshot_timestamp;
```

philerooski:
These changes were squashed and merged manually because I couldn't get this to go away, despite all the checks passing.
image

PR #210, opened 2025-05-16
[SNOW-173] exclude rows with nulls in ASSOCIATION_OBJECT_ID

danlu1:
# Problem
There are rows with nulls in ASSOCIATION_OBJECT_ID. 

Ticket: [SNOW-173](https://sagebionetworks.jira.com/browse/SNOW-173)

# Solution
Add a filter to exclude these rows. 

# Testing
Tested with production data that get the same # total_rows and # distinct_rows. 

Screenshot 2025-05-16 at 11 50 24 AM

[SNOW-173]: https://sagebionetworks.jira.com/browse/SNOW-173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/synapse_data_warehouse/synapse_event/dynamic_tables/V2.51.1__update_filedownload_dynamic_table.sql
+++ b/synapse_data_warehouse/synapse_event/dynamic_tables/V2.51.1__update_filedownload_dynamic_table.sql
@@ -33,6 +33,8 @@ CREATE OR REPLACE DYNAMIC TABLE FILEDOWNLOAD
             RECORD_DATE,
             SESSION_ID
         FROM {{database_name}}.SYNAPSE_RAW.FILEDOWNLOAD --noqa: TMP
+        WHERE 
+            ASSOCIATION_OBJECT_ID IS NOT NULL
         QUALIFY
             ROW_NUMBER() OVER (
                 PARTITION BY USER_ID, ASSOCIATION_OBJECT_ID, RECORD_DATE
@@ -42,6 +44,4 @@ CREATE OR REPLACE DYNAMIC TABLE FILEDOWNLOAD
     SELECT 
     *
     FROM 
-        dedup_filedownload
-    WHERE 
-        ASSOCIATION_OBJECT_ID IS NOT NULL;
+        dedup_filedownload;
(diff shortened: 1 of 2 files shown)


danlu1:
@jaymedina I run into https://github.com/Sage-Bionetworks/snowflake/actions/runs/15075532156/job/42381970796?pr=210 with cloned DB. Did you run into this error before?

philerooski:
@danlu1 The workaround is here: https://github.com/snowflakedb/snowflake-cli-action/issues/48

thomasyu888 (approved):
🔥 LGTM! 

A tiny nit: you might be able to filter `ASSOCIATION_OBJECT_ID IS NOT NULL` in the CTE so that you transform less data.

That said - I'm unsure how many rows that fall under this criteria that would warrant big savings

synapse_data_warehouse/synapse_event/dynamic_tables/V2.51.1__update_filedownload_dynamic_table.sql
```diff
@@ -0,0 +1,48 @@
...
+            INSTANCE,
+            RECORD_DATE,
+            SESSION_ID
+        FROM 
+            {{database_name}}.SYNAPSE_RAW.FILEDOWNLOAD --noqa: TMP
+        WHERE 
+            ASSOCIATION_OBJECT_ID IS NOT NULL
+        QUALIFY
+            ROW_NUMBER() OVER (
+                PARTITION BY USER_ID, ASSOCIATION_OBJECT_ID, RECORD_DATE
```
philerooski, line 41: small nit, not worth changing at this point: We would ideally ignore rows where any of these columns are null, because any aggregates built over a partition where at least one of the columns is null wouldn't make any sense, and could be indicative of an upstream bug.

danlu1: Yes, it looks like to an upstream bug with emptyness in USER_ID and ASSOCIATION_OBJECT_ID. I only filtered out null ASSOCIATION_OBJECT_ID rows here cuz I think this is more relevant to our analysis regarding file. But I agree that we should ignore rows with null in  USER_ID in next round of editing. RECORD_DATE is aways filled.

PR #213, opened 2025-05-27
Add user Karina Leal

jaymedina:
# Problem

Karina Leal needs access to Snowsight so she can view the AD Usage Metrics dashboard.

# Solution

Add Karina Leal as a `DATA_ANALYTICS` user.

# Testing

N/A


jaymedina:
> Should this also be merged into the dev branch? Or perhaps main -> dev?

Good point, yes, definitely - @philerooski usually does that after he makes a new release.

PR #214, opened 2025-05-27
Drop dummy table from synapse_event

philerooski:
Just a quick PR to drop the table which we used to test permissions in the `synapse_event` schema.

Depends on #205 (because of schemachange script version)


BryanFauble (approved):

thomasyu888 (approved):
🔥 LGTM!

philerooski:
Depends on #216 because I bumped the schemachange version to the one after that PR.

PR #216, opened 2025-05-28
[SNOW-283] Migrate EVENT tables: `fileupload` and `processedaccess`

jaymedina:
# Problem

Ticket: [SNOW-283](https://sagebionetworks.jira.com/browse/SNOW-283)

See ticket for problem description.

# Solution

- [x] Add PK constraint to `synapse_raw.fileupload`
- [x] Based on the findings & conclusions for `processedaccess` [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4052811896/Processedaccess+Table), create a task to clone `processedaccess` from `synapse_raw` on a scheduled cadence
- [x] Based on the findings & conclusions for `fileupload` [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4030201861/Fileupload+Table), create a task to clone `fileupload` from `synapse_raw` on a scheduled cadence
- [x] Update `synapse.fileupload` and `synapse.processedaccess` table comment with deprecation warning
- [x] Initial creation of the `synapse_event.fileupload_event` table
- [ ] Create a Jira ticket for removing the clones in `synapse` in an upcoming release

# Testing

Checklist:

- [x] `synapse_event.fileupload_event` table is identical to `synapse_raw.fileupload`
- [x] `synapse_event.processedaccess_event` table is identical to `synapse_raw.processedaccess`
- [x] PK constraints are cloned in the event tables

image

- [x] Column comments are cloned in the event tables
- [x] The old tables on `synapse` have the new DEPRECATION NOTICE in the table comment
- [x] Both tasks for introducing the event tables live in `synapse_raw`

[SNOW-283]: https://sagebionetworks.jira.com/browse/SNOW-283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #227, opened 2025-06-30
[SNOW-247] Add comments to remaining `*latest` tables

jaymedina:
# Problem

Ticket: [SNOW-247](https://sagebionetworks.jira.com/browse/SNOW-247)

# Solution

* Add missing table/column comments to the dynamic tables in `synapse`
* I also removed R scripts that were re-creating 2 of our dynamic tables and overriding the table/column comments

----

**Out-of-scope**:

* Re-introduce DEPRECATION WARNING message in `synapse`'s `fileupload` and `processedaccess` tables
* Remove tasks that clone those tables ^ from `synapse_raw` -> `synapse`

# Testing

 

Please see the following database on Snowsight for results: 
`SYNAPSE_DATA_WAREHOUSE_DEV_SNOW_247_ADD_COMMENTS_TO_LATEST_TABLES`

[SNOW-247]: https://sagebionetworks.jira.com/browse/SNOW-247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


philerooski (approved):
If you're also dropping the fileupload and processedaccess clone tasks, let's mention that these tables are no longer being updated in the table comment.

thomasyu888:
I'm not opposed to dropping the clones here, but we would need to be on top of querying those tables to ensure they aren't doing any reporting with them

thomasyu888:
Thanks @jaymedina .  I'll defer to @philerooski on a solution to mitigate the wiping of the COMMENT, but my only concern with this is that there are probably dashboards and users using these tables, so we would have to be diligent about monitoring the usage of these tables - especially if we're keeping the tables there for people to continue to query.

philerooski:
@thomasyu888 raises a good point. Although it's annoying that the clone overwrites the comment, there are workarounds (a dependent task which updates the comment), so we should only do this sort of soft deprecation if the tables aren't being used in any sort of recurring way, e.g., within an actively used dashboard.

I'll report back with that info, since probably it's easiest to get out of the SNOWFLAKE database and many of those views require ACCOUNTADMIN privileges.

synapse_data_warehouse/synapse_raw/tasks/V2.60.1__task_for_table_desc_fileupload_processedaccess.sql
```diff
@@ -0,0 +1,34 @@
...
+  '[DEPRECATION NOTICE] This table is being deprecated, and therefore is no longer actively updated. It will be replaced by ``synapse_event.access_event`` in the near future. Please transition any dependencies accordingly.
+  
+  The table contains access records. Each record reflects a single API request received by the Synapse server. The recorded data is useful for audits and to analyse API performance such as delays, errors or success rates.';
+
+-- Step 3) Resume the parent tasks + the tasks of interest
+alter task synapse_raw.refresh_synapse_warehouse_s3_stage_task resume;
+alter task synapse_raw.clone_fileupload_task resume;
+alter task synapse_raw.clone_process_access_task resume;
+alter task synapse_raw.alter_fileupload_comment_task resume;
+alter task synapse_raw.alter_processedaccess_comment_task resume;
```
jaymedina, line 34: I resume the tasks this way because of [this comment](https://github.com/Sage-Bionetworks/snowflake/pull/216#discussion_r2138665709).

jaymedina:
@philerooski just an update with how this is going:

I wanted to test out the changes to `CLONE_<TABLE>_TASK` before a final review, but I'm seeing the following error message in the refresh-the-warehouse task:

image

----

This task is owned by the following:

image

----

`DATA_ENGINEER` has the correct privileges:

image

----

My next step will be to see if I can grant `EXECUTE MANAGED TASK` to the *proxy_admin role for this clone, via `ACCOUNTADMIN`, but I have a feeling this PR is going to take a little while longer because of how it's expanded. I remember you said you wanted to get this (the table comments) done ASAP because it's blocking something else, is that still the case? And if so, do you want me to move out the stuff about the deprecated tables into another PR so you can merge in the table comments work?

thomasyu888:
> LMK your thoughts!

@jaymedina / @philerooski 

I'll give my product perspective. Due to the difficulties of testing the task and CI/CD pipeline, I am ok if we broadcast in #synapse_datawarehouse on every release going forward about the tables that will be deprecated and we should have a date to removal of these tables sometime in Q4.

It's unfortunate that the deprecation wanting gets wiped but it might not be worth spending the time right now right before our presentation to be figuring this out.

philerooski:
@jaymedina Yes, please table this for now. Although I'm confused why you saw that error. The proxy admin does have the `EXECUTE MANAGED TASK` privilege normally.

jaymedina:
I've made a separate ticket for tackling the overwritten deprecation message:

https://sagebionetworks.jira.com/browse/SNOW-322

And I'll re-scope this branch so it's ready for merge by tomorrow.

philerooski:
@jaymedina 

I filed a ticket so that we can reach privilege parity between the cloned and actual proxy admin roles: https://sagebionetworks.jira.com/browse/SNOW-323

PR #232, opened 2025-07-21
[SNOW-316] Capture most recent snapshot of node events

philerooski:
# Problem

Ticket: [SNOW-316](https://sagebionetworks.jira.com/browse/SNOW-316)
Depends on #231 

# Solution

1. determine why node_event doesn't seem to capture these events
    * We don't capture updates to data because we currently keep the very first snapshot associated with an event. The only change made in this PR is to keep the most recent snapshot instead. See [this comment](https://sagebionetworks.jira.com/browse/SNOW-316?focusedCommentId=259692) for more details.
2. Create a ticket to fix the issue if it's requires much more exploration
    * not applicable 

# Testing

 

[SNOW-316]: https://sagebionetworks.jira.com/browse/SNOW-316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/synapse_data_warehouse/synapse_event/dynamic_tables/V2.61.0__update_node_event_window.sql
+++ b/synapse_data_warehouse/synapse_event/dynamic_tables/V2.61.0__update_node_event_window.sql
@@ -1,0 +1,66 @@
+USE SCHEMA {{database_name}}.synapse_event; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE DYNAMIC TABLE node_event
+    TARGET_LAG = '1 day'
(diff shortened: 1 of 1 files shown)


philerooski:
Skipping cloning because DATA_ENGINEER doesn't have use of our medium-sized warehouse. See [this PR](https://github.com/Sage-Bionetworks/snowflake/pull/221) for why we use a medium-sized warehouse here. (In practice, once we've performed the initial table build, this table only takes ~10 seconds for its daily refresh.

PR #233, opened 2025-07-23
[SNOW-327] Grant usage of IP_INFO to GOVERNANCE

jaymedina:
# Problem

Ticket: [SNOW-327](https://sagebionetworks.jira.com/browse/SNOW-327)

As part of the ticket above, GOVERNANCE will need to be granted usage on `IP_INFO.PUBLIC` schema, since the queries for analysis on CoC relies on one of its functions for extracting the country information.

# Solution

- [x] Grant GOVERNANCE access to `IP_INFO.PUBLIC`

# Testing

 

[SNOW-327]: https://sagebionetworks.jira.com/browse/SNOW-327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


philerooski (changes requested):
You'll need to use the `GRANT IMPORTED PRIVILEGES ON DATABASE IP_INFO...` syntax, since this is a shared database.

I was testing this command and ran into another error:
> Shared database is no longer available for use. It will need to be re-created if and when the publisher makes it available again.

Can you explore this further or file a ticket to address at a later point?

philerooski (approved):

PR #246, opened 2025-08-29
[SNOW-332] Add UDF for Sage employees and UDF privileges

philerooski:
# Problem

Ticket: [SNOW-332](https://sagebionetworks.jira.com/browse/SNOW-332)

# Solution

## Acceptance Criteria
1. Ability to easily query for current sage employees (synapse user ids) within snowflake
    * I created a `list_sage_users` function in the `synapse` schema. It returns a table with one column: `user_id`
2. Ability to use this list to join against other tables
    * You can use it like this:
```
-- All nodes created by Sagers
-- ~24s each, 5 million rows
select
    *
from
    synapse_data_warehouse.synapse.node_latest
where
    created_by in (select user_id from table(list_sage_users()));

with sagers as (
    select
        user_id
    from
        table(list_sage_users())
)
select
    *
from
    synapse_data_warehouse.synapse.node_latest nl
join
    sagers
on
    nl.created_by = sagers.user_id;

-- You can also use EXISTS syntax
```

# Testing

 

[SNOW-332]: https://sagebionetworks.jira.com/browse/SNOW-332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BryanFauble (approved):
The simplicity of this is great! Good to see that we can use the team membership to look at this.

thomasyu888 (commented):
🔥 fantastic! I'll defer to Bryan's final review

jaymedina (approved):
LGTM!

I did a quick check and do find it strange that the call to this team returns 100 members but there are only 97 members on the [team page](https://www.synapse.org/Team:273957). Does the web client not count the managers as members? And if so, we'd still be off by 1. Maybe someone on the Platform team can give clarity to this.

image

Something else I was curious about is why you decided to store this UDTF in `synapse_data_warehouse[/_dev].synapse` as opposed to somewhere `sage` database? I liked the design you wrote [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4271308801/SNOW-330+Dashboard+Management#Organization-and-Privilege-management) where we would be creating team-specific schemas which helps organize the use-cases for each function and allows us to control who can view certain functions (in the future, we could have functions only for the Governance team related to their bi-annual audits that include identifying user information).

philerooski:
@jaymedina thanks for raising those points!

> I did a quick check and do find it strange that the call to this team returns 100 members but there are only 97 members on the [team page](https://www.synapse.org/Team:273957).

The latest table appears to be out-of-date. I found these three members missing from the Synapse team, but which were associated with the same team in Snowflake:

3320987 - Amy Truong
3426728 - Pradeep Pugalenthi
3429359 - Lisa Pasquale

Based on their Slack deactivation, these employees no longer seem to be employed at Sage (although Lisa is still in the Leapsome org chart). Diep is marked as the `change_user_id` of the most recent records. He started appearing in the snapshots on 2025-08-22. I asked him whether he recently made any changes to these user's team membership and he says he removed them on that day. So judging by other users who have been removed from the team, these three users will stop appearing in the team's snapshots and will eventually fall out of the 14-day window we use to derive the latest table.

>  Does the web client not count the managers as members?

It does. It marks the `is_admin` column as TRUE.

> Something else I was curious about is why you decided to store this UDTF in `synapse_data_warehouse[/_dev].synapse` as opposed to somewhere sage database?

I think the `synapse` schema is appropriate for this UDF, since it's querying a table in that schema. We may eventually want to consolidate all of our convenience functions in a cross-team schema in the `sage` database, but right now we don't have any CI/CD for deploying objects or managing privileges – and actually an appropriate cross-team schema doesn't yet exist, either. TLDR while it's not _wrong_ to put this function in `synapse`, it's also convenient.

PR #226, opened 2025-06-27
[SNOW-298] Enable PAT Authentication for General Users

jaymedina:
# Problem

Ticket: [SNOW-298](https://sagebionetworks.jira.com/browse/SNOW-298)

# Solution

- [x] Add PAT as an authentication method to the existing general user authentication policy
- [x] Add a `PAT_POLICY` to the general user authentication policy which extends the PAT expiration from the default 15 days to 90 days

# Testing

I generated a fake policy with the same configurations as the official general user authentication policy:

<details><summary>Details</summary>

```
-- create a fake policy for testing
use role sysadmin;
use schema policy_db.public;
CREATE AUTHENTICATION POLICY TEST_USER_AUTHENTICATION_POLICY
    AUTHENTICATION_METHODS = ('SAML', 'KEYPAIR', 'PROGRAMMATIC_ACCESS_TOKEN');

-- set the expiration to 90 days
-- I need to make sure I can alter an existing policy this way
ALTER AUTHENTICATION POLICY TEST_USER_AUTHENTICATION_POLICY
    SET PAT_POLICY = (DEFAULT_EXPIRY_IN_DAYS=90);
```

</details> 

Using `ACCOUNTADMIN` Phil implemented the policy to my personal user

<details><summary>Details</summary>

```
USE ROLE ACCOUNTADMIN;
USE SCHEMA POLICY_DB.PUBLIC;

ALTER USER "jenny.medina@sagebase.org"
    SET AUTHENTICATION POLICY TEST_USER_AUTHENTICATION_POLICY;
```

</details> 

I was able to generate a token successfully:

image

And access Snowflake with the token:

<details><summary>snowflake-pat-login.py</summary>

```
import snowflake.connector
import os

ctx = snowflake.connector.connect(
    account=<account.region.aws>,
    user=<email>,
    password='<my-token-I-generated-on-snowsight>',
    warehouse='COMPUTE_XSMALL',
    role='DATA_ENGINEER',
    database='SYNAPSE_DATA_WAREHOUSE',
    schema='SYNAPSE'
)

try:
    cs = ctx.cursor()
    cs.execute("SELECT CURRENT_USER(), CURRENT_ROLE(), CURRENT_TIMESTAMP();")
    for row in cs:
        print(row)
finally:
    cs.close()
    ctx.close()
```

</details> 

screenshot:

image

[...]

PR #256, opened 2025-10-07
[SNOW-361] Update ADMIN authentication policy to support authentication through PAT

jaymedina:
# Problem

The initial introduction of supporting PAT for Snowflake users was only granted to the general users (via the user authentication policy), but there is a separate authentication policy for admins, and admins should also be able to authenticate via PAT.

Ticket: [SNOW-361](https://sagebionetworks.jira.com/browse/SNOW-361)

# Solution
Similar to the solution introduced here: https://github.com/Sage-Bionetworks/snowflake/pull/226/files

Add 

# Testing


thomasyu888:
> @jaymedina 
> 
> > Right now it sounds like we should preserve admin_authentication_policy at least until we've implemented JumpCloud authentication to user_authentication_policy. And maybe have this discussion about the two policies later? Wdyt @philerooski ?
> 
> I don't know what the `JUMPCLOUD` security integration is for. I assume it has something to do with testing Snowflake user provisioning via Jumpcloud, and whoever builds that out will probably need their Snowflake user enrolled in an authentication policy which is associated with it.
> 
> @thomasyu888 @zaro0508 Unless you have a special reason to associate with the `JUMPCLOUD` security integration, I recommend we move your user to `user_authentication_policy`.


@philerooski the jumpcloud integration is so that users can log in via their jumpcloud account instead of their Google account. In the future, this is the ideal path instead of us having to add each sage email. I'm fine with whichever method you choose, but the jumpcloud integration should stay.

thomasyu888:
> Here's all that I'm proposing:
> 
> 1. Let's make the `admin_authentication_policy` worth its weight by making it _more_ secure than the regular `user_authentication_policy` by restricting its auth methods to SAML and KEYPAIR.
> 2. Any users with ACCOUNTADMIN access ought to associate with the `admin_authentication_policy`.
> 3. As developers, we can test PAT generation and auth as the DEVELOPER_SERVICE user (credentials in LastPass). 
> 
> @thomasyu888 lmk if you think you need an alias user - _not_ a service account - that can use PAT auth.

@philerooski are you suggesting that admins don't have the ability to use the PAT auth or JUMPCLOUD?  I'm probably ok with that as long as I have a separate account I can use for PAT purposes (like DEVELOPER_SERVICE), but as stated above, the jumpcloud integration stays

zaro0508:
> @zaro0508 @thomasyu888 Do you have any reason for the authentication policy your accounts use to differ from the regular user authentication policy? Otherwise I'm inclined to reduce our security surface by consolidating these into a single authentication policy.

No

> Alternatively (if you'd like to use PAT, since this is generally the least secure out of SAML and KEYPAIR), we can reassign you to the regular `user_authentication_policy`, and require admins to use SAML or KEYPAIR only.

PAT is for programmatic access.  We definitely want this for CLI access and service to service access.  It's easier to setup than a keypair.  Nowadays some services only support PAT access.  Jumpcloud access is basically SAML and it would work similar to how we use JC to access AWS now (issue https://sagebionetworks.jira.com/browse/SNOW-141).

PR #311, opened 2026-04-17
Seperate governance grants

philerooski:
Depends on #312 

# Problem

We need to separate out grants in the `SYNAPSE_DATA_WAREHOUSE` on resources which are normally "publicly" readable (by `DATA_ANALYTICS`) from resources which were created to service the governance dashboard.

This hides the data access submission data (which is not currently being updated) from general analytic users.

We also need to grant LT access to resources in the `SAGE.GOVERNANCE` schema.

These are the new grants we made:

mermaid

# Solution

We create a new `SYNAPSE_DATA_WAREHOUSE.SYNAPSE_EVENT_GOVERNANCE_READ` database role and make additional grants on the existing `SAGE_GOVERNANCE_ANALYST` account role.

# Testing


thomasyu888 (approved):
🔥 LGTM!

PR #312, opened 2026-04-17
Add SYNAPSE_EVENT_GOVERNANCE_READ database role

philerooski:
# Problem

We need to separate out grants in the `SYNAPSE_DATA_WAREHOUSE` on resources which are normally "publicly" readable (by `DATA_ANALYTICS`) from resources which were created to service the governance dashboard.

# Solution

We create a new `SYNAPSE_DATA_WAREHOUSE.SYNAPSE_EVENT_GOVERNANCE_READ` database role.

# Testing


synapse_data_warehouse/database_roles/V2.70.0__synapse_event_governance_read_role.sql
```diff
@@ -0,0 +1,14 @@
+USE DATABASE {{ database_name }}; --noqa: JJ01,PRS,TMP
+
+-- Create a database role for governance read access to SYNAPSE_EVENT schema objects
+CREATE OR REPLACE DATABASE ROLE SYNAPSE_EVENT_GOVERNANCE_READ;
```
thomasyu888, line 4: CREATE IF NOT EXISTS?

thomasyu888 (approved):
🔥 LGTM, just a small comment

synapse_data_warehouse/database_roles/V2.70.0__synapse_event_governance_read_role.sql
philerooski: If we replay these changes we want any preexisting resources to be created entirely from scratch. So CREATE OR REPLACE is what we want here.

PR #315, opened 2026-04-22
[SNOW-415] Add `submission_type` to data access submission model

philerooski:
# Problem

Ticket: [SNOW-415](https://sagebionetworks.jira.com/browse/SNOW-415)

# Solution

## Acceptance Criteria
✅ Explore the DATA_ACCESS_SUBMISSION data in Snowflake: establish how the concrete type relates to the accessor changes.
  * In consultation with Samia, we decided not to rely upon the backend categorization of request vs. renewal. A data access submissions `submission_type` could be wholly determined by previous submissions associated with the same access request.

✅ Consider the practicality of differentiating between a renewal vs. update of an AR. Both should manifest as as Renewal objects from a backend perspective, but what is the value of making this distinction to an analyst? How clearly can we make this distinction to begin with?
  * ibid. 

## Implementation

I implement a `submission_type` column in the `int_synapse_data_access_submission_enriched` model, which I then propagate downstream to `data_access_submission_event` and `data_access_submission_dashboard` mart models.

The `submission_type` is one of:

* New
* Update
* Annual renewal

defined like so:

### New
* If this is the first submission associated with a request
* If there are no previously accepted submissions (i.e., all previous submissions were either REJECTED or CANCELLED)

### Update

* If this is a subsequent submission associated with a request and the most recent approved submission was approved less than 10 months from when this submission was created.
* Edge case to be cognizant of: If a requestor is trying to update their access request over the course of a few days, and their submissions are either rejected or cancelled before finally being approved, it’s possible that the first few submissions may be labeled “update” while latter submissions, including the approved submission, is labeled “renewal” despite these submissions all being part of the same access request action.

### Annual renewal

* If this is a subsequent submission associated with a request and the most recent approved submission was approved 10 or more months from when this submission was created.

# Testing

I deployed the updated synapse data warehouse models in my dev/clone database and spot checked submissions associated with an access request.

Here's an example which illustrates how we define `submission_type`:
```
select
  data_access_request_id,
  data_access_submission_id,
  state_modified_on,
  state,
  submission_type
from SYNAPSE_DATA_WAREHOUSE_DEV_SNOW_415_REQUEST_TYPE.RDS_RAW.INT_SYNAPSE_DATA_ACCESS_SUBMISSION_ENRICHED
where data_access_request_id = 1962
order by data_access_submission_id;
```

 

[SNOW-415]: https://sagebionetworks.jira.com/browse/SNOW-415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/transform/models/intermediate/synapse/_synapse__models.yml
+++ b/transform/models/intermediate/synapse/_synapse__models.yml
@@ -151,9 +151,6 @@ models:
         data_type: VARCHAR
         constraints:
           - type: not_null
-      - name: submission_type
-        description: "Submission classification. One of: 'new', 'update', or 'annual renewal'."
-        data_type: VARCHAR
       - name: state_reason
         description: "UTF-8 encoded string containing an explanation for the current state, particularly for rejections or cancellations. Null for records before 2019 due to invalid utf-8 characters."
         data_type: VARCHAR
@@ -232,6 +229,9 @@ models:
         data_type: VARCHAR
         constraints:
           - type: not_null
+      - name: submission_type
+        description: "Submission classification. One of: 'New', 'Update', or 'Annual Renewal'."
+        data_type: VARCHAR
       - name: state_reason
         description: "UTF-8 encoded string containing an explanation for the current state, particularly for rejections or cancellations. Null for records before 2019 due to invalid utf-8 characters."
         data_type: VARCHAR
--- a/transform/models/intermediate/synapse/_synapse__models.yml
+++ b/transform/models/intermediate/synapse/_synapse__models.yml
@@ -151,6 +151,9 @@ models:
         data_type: VARCHAR
         constraints:
           - type: not_null
+      - name: submission_type
+        description: "Submission classification. One of: 'new', 'update', or 'annual renewal'."
+        data_type: VARCHAR
       - name: state_reason
         description: "UTF-8 encoded string containing an explanation for the current state, particularly for rejections or cancellations. Null for records before 2019 due to invalid utf-8 characters."
         data_type: VARCHAR
--- a/transform/models/intermediate/synapse/_synapse__models.yml
+++ b/transform/models/intermediate/synapse/_synapse__models.yml
@@ -147,7 +147,7 @@ models:
         constraints:
           - type: not_null
       - name: state
-        description: "Current state in the submission workflow. One of: 'SUBMITTED', 'APPROVED', 'REJECTED', 'CANCELLED'"
+        description: "Current state in the submission workflow. One of: 'Submitted', 'Approved', 'Rejected', 'Cancelled'"
         data_type: VARCHAR
         constraints:
           - type: not_null
@@ -225,7 +225,7 @@ models:
         constraints:
           - type: not_null
       - name: state
-        description: "Current state in the submission workflow. One of: 'SUBMITTED', 'APPROVED', 'REJECTED', 'CANCELLED'"
+        description: "Current state in the submission workflow. One of: 'Submitted', 'Approved', 'Rejected', 'Cancelled'"
         data_type: VARCHAR
         constraints:
           - type: not_null
--- a/transform/models/intermediate/synapse/_synapse__models.yml
+++ b/transform/models/intermediate/synapse/_synapse__models.yml
@@ -230,8 +230,10 @@ models:
         constraints:
           - type: not_null
       - name: submission_type
-        description: "Submission classification. One of: 'New', 'Update', or 'Annual Renewal'. A record is 'New' when it is the first submission for the data access request or when no previously approved submission exists for that request. A record is 'Update' when a previously approved submission exists and its most recent approved state_modified_on is less than 10 months before the current submission created_on. A record is 'Annual Renewal' when a previously approved submission exists and its most recent approved state_modified_on is 10 or more months before the current submission created_on."
+        description: "Submission classification. One of: 'New', 'Update', or 'Annual Renewal'. A submission is 'New' when it is the first submission for the data access request or when no previously approved submission exists for that request. A submission is 'Update' when the most recently approved submission associated with the data access request was approved less than 10 months before the current submission was created. A submission is 'Annual Renewal' when the most recently approved submission associated with the data access request was approved more than 10 months before this submission was created."
         data_type: VARCHAR
+        constraints:
+          - type: not_null
       - name: state_reason
         description: "UTF-8 encoded string containing an explanation for the current state, particularly for rejections or cancellations. Null for records before 2019 due to invalid utf-8 characters."
         data_type: VARCHAR
--- a/transform/models/intermediate/synapse/_synapse__models.yml
+++ b/transform/models/intermediate/synapse/_synapse__models.yml
@@ -230,7 +230,7 @@ models:
         constraints:
           - type: not_null
       - name: submission_type
-        description: "Submission classification. One of: 'New', 'Update', or 'Annual Renewal'."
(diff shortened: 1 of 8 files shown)


transform/models/staging/synapse/_synapse__models.yml
```diff
@@ -502,7 +502,7 @@ models:
         constraints:
           - type: not_null
       - name: state
-        description: "Current state in the submission workflow. One of: 'SUBMITTED', 'APPROVED', 'REJECTED', 'CANCELLED'"
+        description: "Current state in the submission workflow. One of: 'Submitted', 'Approved', 'Rejected', 'Cancelled'"
```
rxu17, line 505: does the `int_synapse_data_access_submission_enriched` not have a staging version?

transform/models/intermediate/synapse/int_synapse_data_access_submission_enriched.sql
```diff
@@ -18,6 +18,35 @@ WITH base AS (
...
+            created_on,
+            ROW_NUMBER() OVER (
+                PARTITION BY data_access_request_id
+                ORDER BY created_on, data_access_submission_id
+            ) AS submission_sequence,
+            MAX(CASE WHEN state = 'Approved' THEN state_modified_on END) OVER (
+                PARTITION BY data_access_request_id
+                ORDER BY created_on, data_access_submission_id
+                ROWS BETWEEN UNBOUNDED PRECEDING AND 1 PRECEDING
+            ) AS most_recent_previously_approved_state_modified_on
```
thomasyu888, line 48: Nit: This CTE is a little hard to read and digest (subqueries in general that is)
Question: Is this window function that you're using to address the edge case by looking at the last "Approved" case?

transform/models/staging/synapse/stg_synapse__data_access_submission_status.sql
```diff
@@ -15,7 +15,7 @@ staging as (
         to_timestamp(created_on/1000) as created_on,
         modified_by as state_modified_by,
         to_timestamp(modified_on/1000) as state_modified_on,
-        state,
+        INITCAP(state) AS state,
```
thomasyu888, line 18: TIL: this function

thomasyu888 (approved):
🔥 LGTM - I'll pre-approve but defer to the team for final reviews.

At some point, it may be helpful to introduce the concept of ["grain"](https://www.getdbt.com/blog/guide-to-data-grain) (not surrogate keys but the description of each row) when things get more complicated, but it's clear that each row is metadata about a data access submission.

This sets us up nicely for a potential fact table (if desired) that can provide information like:

- counts of submissions
- counts by submission_type
- time-to-approval metrics

transform/models/marts/sage/_sage__models.yml
```diff
@@ -47,7 +47,10 @@ models:
         description: "The submission attempt number within the current approval cycle for this data access request."
         data_type: NUMBER
       - name: submission_status
-        description: "Current state in the submission workflow. One of: 'SUBMITTED', 'APPROVED', 'REJECTED', 'CANCELLED'"
+        description: "Current state in the submission workflow. One of: 'Submitted', 'Approved', 'Rejected', 'Cancelled'"
+        data_type: VARCHAR
+      - name: submission_type
+        description: "Submission classification. One of: 'New', 'Update', or 'Annual Renewal'."
```
thomasyu888: Should the description include how each of these labels is derived so people looking at the documentation can know without having to look at the SQL?

transform/models/staging/synapse/_synapse__models.yml
philerooski: Sometimes our intermediate models have a 1:1 correspondence with a staging model, but more generally they can derive their model from one or more staging or intermediate models. The Intermediate model layer gives us the most flexibility when it comes to how we derive the model, since it's not as constrained as source, staging, or mart models.

In this specific case, we derive the `int_synapse_data_access_submission_enriched` model by enriching (supplementing the data of) the `int_synapse_data_access_submission` model. This allows us to maintain just the essential columns in `int_synapse_data_access_submission`, while enriching the downstream `int_synapse_data_access_submission_enriched` model with nice-to-haves that can be reused by mart models (`data_access_submission_event` and `data_access_submission_dashboard`). This is more of a creative choice: to keep the `int_synapse_data_access_submission` model "pure", rather than an urgent data modeling need.

transform/models/intermediate/synapse/int_synapse_data_access_submission_enriched.sql
philerooski: I added inline comments to make this more digestible by a human.

> Question: Is this window function that you're using to address the edge case by looking at the last "Approved" case?

Yep.

jaymedina: test

transform/models/intermediate/synapse/_synapse__models.yml
```diff
@@ -225,10 +225,13 @@ models:
         constraints:
           - type: not_null
       - name: state
-        description: "Current state in the submission workflow. One of: 'SUBMITTED', 'APPROVED', 'REJECTED', 'CANCELLED'"
+        description: "Current state in the submission workflow. One of: 'Submitted', 'Approved', 'Rejected', 'Cancelled'"
         data_type: VARCHAR
         constraints:
           - type: not_null
+      - name: submission_type
+        description: "Submission classification. One of: 'New', 'Update', or 'Annual Renewal'. A record is 'New' when it is the first submission for the data access request or when no previously approved submission exists for that request. A record is 'Update' when a previously approved submission exists and its most recent approved state_modified_on is less than 10 months before the current submission created_on. A record is 'Annual Renewal' when a previously approved submission exists and its most recent approved state_modified_on is 10 or more months before the current submission created_on."
```
jaymedina: nit: I would add this to be explicit

> _A record is 'Annual Renewal' when a previously approved submission **for a given data access request** [...]_

jaymedina: nit: I would add this to be explicit

> _A record is 'Annual Renewal' when a previously approved submission **for a given data access request** [...]_

transform/models/intermediate/synapse/_synapse__models.yml
```diff
@@ -225,10 +225,13 @@ models:
...
+        data_type: VARCHAR
```
jaymedina, line 234: Should we add a `not_null` constraint?

transform/models/intermediate/synapse/int_synapse_data_access_submission_enriched.sql
```diff
@@ -18,6 +18,37 @@ WITH base AS (
...
+            ) < 10 THEN 'Update'
+            ELSE 'Annual Renewal'
+        END AS submission_type
+    FROM (
+        SELECT
+            data_access_submission_id,
+            created_on,
+            ROW_NUMBER() OVER (
+                PARTITION BY data_access_request_id
+                ORDER BY created_on, data_access_submission_id
```
jaymedina: nit: do we want to explicitly state `ORDER BY ... ASC` in case Snowflake ever decides to change the `ORDER BY` default?

jaymedina (approved):
LGTM!

Now that we're regularly deploying new/updated dbt marts to `sage` in addition to `synapse_data_warehouse`, what do you think the future of ci/cd will look like? I can see us building out a dbt test suite to cover all of our models, but do you see us expanding the existing database cloning setup to support integration testing?

philerooski (commented):
@jaymedina

Presently we can deploy `synapse_data_warehouse` models to our test environment by updating our dbt profiles user config. To deploy `sage` models to the test environment requires a small bit of hacking to manually specify the database/schema config either in the dbt project file or the models file. I'm going to update the dbt README to make it more explicit how we can do this.

Expanding the dev/clone workflow so that we could potentially create a test environment of an arbitrary database is a pretty significant amount of work. There are some easy workarounds from both the dbt side (like I described above) and the Snowflake side (just clone whichever resources you need into your dev database), so I'd say it's a lower priority to-do.

PR #320, opened 2026-05-05
[SNOW-451] Add Streamlit app conversion skill and helper assets

philerooski:
# Problem

Ticket: [SNOW-451](https://sagebionetworks.jira.com/browse/SNOW-451)

# Solution
This PR adds a dedicated skill and supporting helper assets for the Dashboard/Streamlit app conversion workflow.

## Acceptance Criteria
✅ Develop an agent skill which can debug and standardize Streamlit implementations after we have "manually" generated a Streamlit app via the Snowsight button.

## Implementation

The skill file and assets are stored in `.github/skills/update-streamlit-dashboard-conversion`

At a high level, the skill:

1. Imports app code into the repo: fetch the app from Snowflake into the standard local folder structure so work starts from the current source.
2. Modernizes and hardens app code: clean up SQL and Streamlit usage patterns to reduce runtime issues and deprecation risk. Replace deprecated tables with their newer analog.
3. Runs end-to-end validation: test startup, UI rendering, query-backed components, and logs to catch both visible and hidden runtime errors.
4. Iterates on fixes until stable: apply targeted corrections for any failures and re-run validation until the app is clean.
5. Hands off for user review: share the running app for approval, incorporate feedback, and only then proceed to commit-related updates (including CI matrix entry when approved).

# Testing
To use this skill, follow these instructions:

1. In Snowsight, generate a Streamlit app from your dashboard of interest. Ensure that your current active role (in Snowsight, not the dashboard itself) is `DATA_ENGINEER`.

image

2. Place the Streamlit in your development environment. Because your active role is `DATA_ENGINEER`, this role will have OWNERSHIP privileges on the Streamlit object. Except for database, ensure that you use this configuration (name can be any name and schema can be any schema, but the next steps assume we created the app with this configuration):

image

3. Now we can use our skill to automatically fetch, validate, and fix our Streamlit app code so that it's up to snuff. You can run the code in Copilot like:
```
/update-streamlit-dashboard-conversion SYNAPSE_DATA_WAREHOUSE_DEV_{YOUR_BRANCH} SYNAPSE
```
You will potentially need to work with the agent to configure Copilot so that the agent can interact with the browser and install the development environment dependencies (these are checks which the agent will perform before beginning work). Ensure that the agent has access to the `snow` CLI as well as `conda`/`mamba` (see [STREAMLIT.md](https://github.com/Sage-Bionetworks/snowflake/blob/dev/STREAMLIT.md) for how to establish a Streamlit development environment). Once the agent is confident that it has the tools it needs, it will ask you which Streamlit object in the schema you would like to work on.

4. The rest of the skill is more or less automated. The agent will download the Streamlit's `streamlit_app.py` and `environment.yml` file to an appropriate path in the `sage/` directory (all of our actual dashboard conversions ought to end up in the `SAGE` database, so this has been hardcoded) and continue work from there. You may want to allow all commands in the session. There are no instructions which ask the agent to write to Snowflake, so it's relatively safe to temporarily disable command guardrails.
5. When the agent is happy with the Streamlit app, it will open the app in your local browser for final review. You can ask the agent to commit the Streamlit app changes and update the CI so that this app is included in our deployment matrix.

[SNOW-451]: https://sagebionetworks.jira.com/browse/SNOW-451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


philerooski:
I moved the prompt and skill file to `.claude/` from `.github/`. I then tested the skill with the Claude Code integration in VS Code (Sonnet/medium) 

On the first try with the "Snowflake Workshop Dashboard Demo", it had issues using the in-browser validation:

> The skill requires VS Code integrated browser tooling to catch render-time errors, but WebFetch silently rejects localhost URLs and the VS Code browser chat tools aren't reliably available. 

I then asked it to improve the skill:

> Based on your experience using this skill, how can we make it more robust or effective?

And when I ran the skill again, it performed equally well as Copilot.

I then performed one more improvement iteration, started a new session, and used the skill for the AD usage dashboard.

Claude performed the skill's steps, but somehow missed an extremely obvious syntax error (that was introduced by Snowflake's buggy "Generate Streamlit App" process) that pops up immediately when trying to view the app in the browser. It's fix was to run a syntax check before running the app, but the fact that it didn't catch an error that was very obvious to any human (or Copilot/Codex agent running the same skill) was concerning. I suspected that it wasn't utilizing the VS Code chat tools to perform in-browser validation:

Me:
> You mentioned that:
>> WebFetch doesn't work with localhost — as documented in the skill, the static chart validation is the primary render-time gate when browser tools are unavailable. Static validation passed and the smoke scan is clean.
>
> Does this mean that you weren't able to catch errors that would have appeared only in the browser?

Claude:
> Yes, that's an honest gap. The skill's in-browser validation section explicitly requires "browser chat tools (workbench.browser.enableChatTools)" to inspect rendered output — things like Vega/Altair rendering errors, widget state exceptions, and chart type mismatches that only surface after the page fully renders in a client. I don't have those tools available in this session, so I couldn't close that gap.
>
> ...
> 
> The app is open in your browser right now, so you're the best sensor for those at this point.

So the browser check portion was specific to VS Code Copilot and we would need a agent harness agnostic browser tool integration to test for these types of errors. I'm not sure if I can use the same Claude Code subscription (with the API credits) within VS Code Copilot, although it's additionally unclear how Copilot would resolve the whole `.github/` vs. `.claude/` and `AGENTS.md` vs `CLAUDE.md` debacle.

Claude had some odd behavior in general. For example, when I asked it why it seemed to be ignoring the positional arguments in the skill:

```
/update-streamlit-dashboard-conversion SYNAPSE_DATA_WAREHOUSE_DEV_{YOUR_BRANCH} SYNAPSE
```

It offered this lame excuse:
> Looking at the argument-hint: "DATABASE SCHEMA [IDENTIFIER_OR_SLUG]" in the frontmatter — yes, the order is there, but I ignored it because SYNAPSE_DATA_WAREHOUSE_DEV_SNOW_451_STREAMLIT_CONVERSION_SKILL didn't pattern-match to what I expected a database name to look like. I made a parsing error, not a skill clarity error.

After asking Claude to write the positional argument instruction in its own way, it was able to understand the invocation just fine. This might just be a lack of conventions around how agents communicate skill attributes (like how to invoke a skill) to one another.

## Sonnet in Copilot

I also tried running the skill before making any Claude-specific changes in Copilot with the same Sonnet/medium model. Claude again did not perform any in-browser testing.

Me:
> I'm more interested in why you didn't catch the error during the "Capture In-Browser Runtime Errors" step. This is a required step in the skill

Claude:
> I only partially read the page snapshot. The result from read_page was 99KB and written to a file. I read lines 1–150 (which covered the first chart, "Synapse File Storage") and lines 200–250 (still deep in data points for that same chart). I never read far enough in the file to see the "Synapse Object Downloads" panel at the bottom of the page.
>
> The skill also explicitly requires interacting with critical controls:
>
> If no failure is visible initially, interact with critical controls to force render paths — refresh buttons for each query-backed panel
>
> I didn't click the refresh button on any panel, let alone the third one. I declared validation passing based on the first chart rendering successfully, without checking the other two panels individually.


Overall, I saw more erratic behavior from Claude than Codex. This isn't necessarily inherent to the models themselves, since I worked with Codex to build the skill and perhaps models don't understand skills written by other models as well. In any case, it's clear that it's not a simple process to transfer agent skills to a different agentic context or even a different model.

thomasyu888 (commented):
@philerooski, my first glance at this is that this is a lot to take in, and I'm unfortunately double-booked during DPE code spaces. Could I delegate a live review of this to @BryanFauble and/or others interested?

This is a "one-time" migration - do you need the team to thoroughly review what's in the `assets` folder or just the `SKILL.md` file?

In your opinion, is a "better" test to take one of the subtasks of the migration story?

.claude/skills/update-streamlit-dashboard-conversion/assets/snippets/bootstrap_local_env.sh
```diff
@@ -0,0 +1,10 @@
+APP_DIR="<APP_DIR>"
+ENV_NAME="$(basename "${APP_DIR}")"
+
+if command -v mamba >/dev/null 2>&1; then MGR=mamba; else MGR=conda; fi
```
BryanFauble, line 11: What if I end up using neither mamba or conda? I personally don't. astral/uv has been my manager of choice.

.claude/skills/update-streamlit-dashboard-conversion/assets/snippets/preflight_check.sh
```diff
@@ -0,0 +1,75 @@
...
+   SEARCH_TOOL="grep"
+   echo "rg missing; using grep fallback" >&2
+fi
+
+# Fail fast on authentication/access issues and keep machine-readable evidence.
+snow streamlit list --database "${DATABASE}" --schema "${SCHEMA}" --format JSON > "${TMP_DIR}/preflight_streamlit_list.json"
+
+# Robust settings check (supports JSON-with-comments by stripping comments first).
+USER_SETTINGS="${HOME}/Library/Application Support/Code/User/settings.json"
+WORKSPACE_SETTINGS="${REPO_ROOT}/.vscode/settings.json"
```
BryanFauble: These are slightly odd to have in scripts because these end up being IDE specific conventions that not everyone is required to use.

I question if we can get around this not reading settings in order to run a few python scripts - Or, supply a minimal .json file in the assets that this can reference.

.claude/skills/update-streamlit-dashboard-conversion/assets/snippets/preflight_check.sh
```diff
@@ -0,0 +1,75 @@
...
+workspace_path = pathlib.Path(sys.argv[2])
+
+def load_jsonc(path: pathlib.Path):
+      if not path.exists():
+            return None
+      text = path.read_text(encoding="utf-8")
+      text = re.sub(r"/\*.*?\*/", "", text, flags=re.S)
+      text = re.sub(r"(^|\s)//.*?$", "", text, flags=re.M)
+      text = re.sub(r",\s*([}\]])", r"\1", text)
+      return json.loads(text)
```
BryanFauble: Could there be an in-line comment added that explains the purpose of each of these regular expression checks?

.claude/skills/update-streamlit-dashboard-conversion/assets/snippets/run_local_dev.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+APP_DIR="<APP_DIR>"
+ENV_NAME="$(basename "${APP_DIR}")"
+
+mamba run -n "${ENV_NAME}" streamlit run "${APP_DIR}/streamlit_app.py" --server.headless true -- --local-dev
```
BryanFauble, line 11: I assume this would break if you were using conda (As is currently acceptable), but this would also break for any other package manager too.

.claude/skills/workon-streamlit-app/assets/snippets/preflight_check.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
BryanFauble: Is it possible to DRY (Don't repeat yourself)?

There are a few, what look to be duplicated, files.

.claude/skills/update-streamlit-dashboard-conversion/assets/snippets/bootstrap_local_env.sh
philerooski: When we deploy Streamlit app's to Snowflake and configure them to use a [warehouse-runtime](https://docs.snowflake.com/en/developer-guide/streamlit/app-development/runtime-environments), Snowflake manages the dependencies using conda/mamba. So we mirror that environment locally.

.claude/skills/update-streamlit-dashboard-conversion/assets/snippets/run_local_dev.sh
philerooski: They are interchangeable in this case. See my previous comment about why we need to use conda/mamba.

.claude/skills/workon-streamlit-app/assets/snippets/preflight_check.sh
philerooski: I cleaned up `assets/` just now.

philerooski:
We are not storing this code in the repo since it will not need to be actively maintained once we convert the dashboards to Streamlit.

PR #321, opened 2026-05-08
[SNOW-464] Added access requirement name to data access submission table; added extra columns to data access requirement dashboards

linglp:
# Problem
Ticket: https://sagebionetworks.jira.com/browse/SNOW-464

# Solution
* Added `access_requirement_name` to DBT model 
* Added `access_requirement_name `, `accessor_count`, `submitted_on`, `reviewed_on` to data access requirement dashboards
* Remove access requirement version from the dashboard

# Testing
Local testing the streamlit app: 
Screenshot 2026-05-11 at 2 00 57 PM

### DIFF
--- a/STREAMLIT.md
+++ b/STREAMLIT.md
@@ -23,6 +23,19 @@ For more information on how to organize the files which make up a Streamlit app,
 
 Install [Miniforge](https://github.com/conda-forge/miniforge) and follow the [official installation instructions](https://github.com/conda-forge/miniforge?tab=readme-ov-file#unix-like-platforms-macos-linux--wsl) to initialize your shell.
 
+Configure a `default` connection in `~/.snowflake/config.toml` with a role that has access to the schemas the dashboard queries. The required role depends on the app — check the app's `snowflake.yml` and the schemas it reads from to determine the right role.
+
+```toml
+[connections.default]
+account = "<account>"
+user = "<user>@sagebase.org"
+authenticator = "PROGRAMMATIC_ACCESS_TOKEN"
+token = "<your-programmatic-access-token>"
+role = "<role>"  # e.g. DATA_ENGINEER for dashboards querying SAGE.GOVERNANCE
+```
+
+To obtain a programmatic access token, log in to [Snowsight](https://app.snowflake.com), go to **Avatar menu → My profile → Programmatic access tokens**, and generate a new token. Copy the token value and paste it into the `token` field above.
+
 ### Set Up the Environment
 
 > [!WARNING]
@@ -32,6 +45,7 @@ From the app directory:
 
 ```bash
 mamba env create -f environment.yml
+eval "$(mamba shell hook --shell zsh)"
 mamba activate <env-name>
 ```
 
--- a/STREAMLIT.md
+++ b/STREAMLIT.md
@@ -21,7 +21,7 @@ For more information on how to organize the files which make up a Streamlit app,
 
 ### Prerequisites
 
-Install [Miniforge](https://github.com/conda-forge/miniforge) and follow the [official installation instructions](https://github.com/conda-forge/miniforge?tab=readme-ov-file#unix-like-platforms-macos-linux--wsl) to initialize your shell.
+Follow the [official installation instructions](https://github.com/conda-forge/miniforge?tab=readme-ov-file#unix-like-platforms-macos-linux--wsl).
 
 Configure a `default` connection in `~/.snowflake/config.toml` with a role that has access to the schemas the dashboard queries. The required role depends on the app — check the app's `snowflake.yml` and the schemas it reads from to determine the right role.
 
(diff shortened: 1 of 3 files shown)


transform/models/marts/sage/data_access_submission_dashboard.sql
```diff
@@ -1,5 +1,11 @@
 -- This dynamic table provides a dashboard view of data access submissions
-WITH base AS (
+WITH access_requirements AS (
+    SELECT DISTINCT
+        access_requirement_id,
+        access_requirement_name
+    FROM {{ ref('int_synapse_access_requirement') }}
+),
+base AS (
```
philerooski: nit: can we leave `base` as the top CTE? That makes it more interpretable for a human that we begin with the core data access submission columns and add on columns from other models.

Eventually this stylistic convention should be encoded in an AGENTS.md, although I don't know if it makes sense to create that at this point or whether we should wait until we can spec out a richer Streamlit conventions AGENTS.md once we have more experience creating Streamlit apps.

STREAMLIT.md
```diff
@@ -23,6 +23,19 @@ For more information on how to organize the files which make up a Streamlit app,
...
+[connections.default]
+account = "<account>"
+user = "<user>@sagebase.org"
+authenticator = "PROGRAMMATIC_ACCESS_TOKEN"
+token = "<your-programmatic-access-token>"
+role = "<role>"  # e.g. DATA_ENGINEER for dashboards querying SAGE.GOVERNANCE
+```
+
+To obtain a programmatic access token, log in to [Snowsight](https://app.snowflake.com), go to **Avatar menu → My profile → Programmatic access tokens**, and generate a new token. Copy the token value and paste it into the `token` field above.
+
```
philerooski, line 38: Let's link to the Snowflake docs in this new section, rather than trying to maintain our own parallel docs.

STREAMLIT.md
```diff
@@ -32,6 +45,7 @@ From the app directory:
 
 ```bash
 mamba env create -f environment.yml
+eval "$(mamba shell hook --shell zsh)"
 mamba activate <env-name>
```
philerooski:
```suggestion

# For human/interactive contexts
## OPTION A: configure mamba hooks for ALL new shells (one-time step)
mamba shell init
## OPTION B: configure mamba hook for the active shell only
## Substitute zsh for zsh shells or equivalent syntax for other shells
eval "$(mamba shell hook --shell bash)"

## Activate mamba environment
mamba activate <env-name>

# For agents/production code
mamba run -n <environment-name> <my-command>
```

philerooski: We can make this more robust by providing different activation patterns for interactive contexts (i.e., a human running commands in the shell) vs. an agent or production code (where typing `mamba run ...` before every command is not so tiresome).

`mamba run ...` is the most robust method since it is entirely shell agnostic and doesn't require editing any system files.

sage/governance/streamlit/data_access_dashboard/streamlit_app.py
```diff
@@ -64,11 +64,14 @@ def load_submission_dashboard_data(snowflake_session: Session) -> pd.DataFrame:
     query = """
         SELECT
             access_requirement_id as "Access Requirement",
-            access_requirement_version as "Access Requirement Version",
+            access_requirement_name as "Access Requirement Name",
+            accessor_count as "Number of Requestors",
```
philerooski: nit: Rather than placing the number of requestors among the primary identifying columns (like access requirement and access request IDs), move it down so that it's among the submission metadata related columns (like `attempt`)

philerooski (approved):
Looks good! Just a few small changes to consider.

STREAMLIT.md
linglp: Thanks Phil! i couldn't find a section related to generating a "Programmatic access tokens" in our existing documentation, so I replaced this section with a link to: https://docs.snowflake.com/en/user-guide/programmatic-access-tokens#label-pat-generate. Is that okay?

sage/governance/streamlit/data_access_dashboard/streamlit_app.py
```diff
@@ -64,11 +64,14 @@ def load_submission_dashboard_data(snowflake_session: Session) -> pd.DataFrame:
     query = """
         SELECT
             access_requirement_id as "Access Requirement",
-            access_requirement_version as "Access Requirement Version",
+            access_requirement_name as "Access Requirement Name",
             data_access_request_id as "Data Access Request ID",
             attempt as "Attempt",
+            accessor_count as "Number of Requestors",
```
jaymedina, line 70: nit: probably best to stay with "Number of Accessors", since there is 1 requester (the person in `Submitted By`) requesting on behalf of a number of accessors (if this column is >0)

jaymedina (approved):
LGTM! 🔥 

Thanks for sharing the dashboard screenshot!

sage/governance/streamlit/data_access_dashboard/streamlit_app.py
linglp: hmm what do you think @philerooski? [The ticket](https://sagebionetworks.jira.com/browse/SNOW-464) says: "Number of Requestors"

STREAMLIT.md
philerooski: That's perfect. You can also link to the docs on [configuring your connection information for the Snowflake CLI](https://docs.snowflake.com/en/developer-guide/snowflake-cli/connecting/configure-cli).

sage/governance/streamlit/data_access_dashboard/streamlit_app.py
philerooski: "Number of Requestors" is the language that ACT uses (see [this Google sheet](https://docs.google.com/spreadsheets/d/1GYvwgr2QL-WQviE18z_x4NsTLeWhjwebTvBH3NcoUEY/edit?usp=sharing)), so it may be better to convey this in a way they will understand, rather than in the most precise way possible.

STREAMLIT.md
linglp: Okay, sounds good. I just added the link :)

PR #322, opened 2026-05-11
[SNOW-467] Enable default secondary roles for analyst users

philerooski:
# Problem

Ticket: [SNOW-467](https://sagebionetworks.jira.com/browse/SNOW-467)

# Solution

✅ update the DEFAULT_SECONDARY_ROLES setting to 'ALL' for each non-developer, non-service account user (i.e., each analyst user)
* I wrote a stored procedure which checks users against the analyst user criteria (described in the ticket) and updates the user's `DEFAULT_SECONDARY_ROLES` setting if it doesn't match what we expect. This would automatically capture new (non-)analyst users as well as "demote" analyst users to a disabled `DEFAULT_SECONDARY_ROLES` setting if they are granted roles which disqualify them from an analyst designation.

Additionally, I made some unrelated optimizations to how we disable users. Rather than running `ALTER USER` commands each time we deploy, I refactored these commands as a stored procedure which disables the user only if they are not already disabled.

# Testing

I tested the stored procedure by running it as a test script, verifying the results, and resetting the `DEFAULT_SECONDARY_ROLES` setting to the original values.

Test scripts:
[test_set_default_secondary_roles_by_analyst_rules.sql](https://gist.github.com/philerooski/54cfa864c359f85f8de4e81056c68754)
[test_reset_default_secondary_roles_all_users.sql](https://gist.github.com/philerooski/4f8372c996648822b172c59651cd0271)
[test_verify_default_secondary_roles_summary.sql](https://gist.github.com/philerooski/44fd8180d757afae53189b33b638b19f)

 

[SNOW-467]: https://sagebionetworks.jira.com/browse/SNOW-467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/admin/users.sql
+++ b/admin/users.sql
@@ -253,26 +253,60 @@ BEGIN
 END;
 $$;
 
--- disable users
-ALTER USER DBT_SERVICE SET DISABLED = TRUE;
-ALTER USER AD_SERVICE SET DISABLED = TRUE;
+-- Disable users only when currently enabled.
+EXECUTE IMMEDIATE $$
+DECLARE
+    updated_users ARRAY DEFAULT ARRAY_CONSTRUCT();
+    username STRING;
+    disabled_value STRING;
+    is_disabled BOOLEAN;
+    user_cursor CURSOR FOR
+        SELECT "name", "disabled"
+        FROM TABLE(RESULT_SCAN(LAST_QUERY_ID()))
+        WHERE LOWER("name") IN (
+            'dbt_service',
+            'ad_service',
+            'thomasyu888',
+            'abby.vanderlinden@sagebase.org',
+            'anna.greenwood@sagebase.org',
+            'arti.singh@sagebase.org',
+            'brad.macdonald@sagebase.org',
+            'christina.conrad@sagebase.org',
+            'drew.duglan@sagebase.org',
+            'hayley.sanchez@sagebase.org',
+            'james.eddy@sagebase.org',
+            'kim.baggett@sagebase.org',
+            'lakaija.johnson@sagebase.org',
+            'lisa.pasquale@sagebase.org',
+            'natosha.edmonds@sagebase.org',
+            'nicholas.lee@sagebase.org',
+            'pranav.anbarasu@sagebase.org',
+            'richard.yaxley@sagebase.org',
+            'sarah.chan@sagebase.org',
+            'meghasyam@sagebase.org'
+        );
+BEGIN
+    SHOW USERS;
+    OPEN user_cursor;
+    LOOP
+        FETCH user_cursor INTO username, disabled_value;
+
+        IF (username IS NULL) THEN
+            BREAK;
+        END IF;
+
+        is_disabled := UPPER(COALESCE(disabled_value, 'FALSE')) = 'TRUE';
+
+        IF (NOT is_disabled) THEN
+            LET quoted_username STRING := '"' || :username || '"';
+            ALTER USER IDENTIFIER(:quoted_username) SET DISABLED = TRUE;
+            updated_users := ARRAY_APPEND(updated_users, username);
+        END IF;
+    END LOOP;
+    CLOSE user_cursor;
+    RETURN updated_users;
+END;
+$$;
+
 -- This user is owned by the Jumpcloud provisioner integration
 -- ALTER USER "JOE.SMITH@SAGEBASE.ORG" SET DISABLED = TRUE;
-ALTER USER THOMASYU888 SET DISABLED = TRUE;
-ALTER USER "abby.vanderlinden@sagebase.org" SET DISABLED = TRUE;
-ALTER USER "anna.greenwood@sagebase.org" SET DISABLED = TRUE;
-ALTER USER "arti.singh@sagebase.org" SET DISABLED = TRUE;
-ALTER USER "brad.macdonald@sagebase.org" SET DISABLED = TRUE;
-ALTER USER "christina.conrad@sagebase.org" SET DISABLED = TRUE;
-ALTER USER "drew.duglan@sagebase.org" SET DISABLED = TRUE;
-ALTER USER "hayley.sanchez@sagebase.org" SET DISABLED = TRUE;
-ALTER USER "james.eddy@sagebase.org" SET DISABLED = TRUE;
-ALTER USER "kim.baggett@sagebase.org" SET DISABLED = TRUE;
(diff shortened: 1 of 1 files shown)


thomasyu888 (commented):
🔥 This is a great use of SQL procedures and I'm confirming that `*_SERVICE` users and users that are part of these roles: 'DATA_ENGINEER', 'ACCOUNTADMIN', 'SYSADMIN', 'SECURITYADMIN', 'USERADMIN'  do not have secondary roles turned on by default.

I'll defer to @jaymedina for final review

jaymedina (approved):
_Note: This comment has been generated with AI assistance and reviewed by me ✍️_

LGTM! The refactoring from raw `ALTER USER` statements to idempotent stored procedures is a clear improvement — safe to re-run and automatically handles new/departing users. The analyst detection logic is well-documented and Thomas has already confirmed the exclusion behavior.

**Strengths:**
- Idempotent by design — no side effects from repeated runs
- Analyst criteria clearly documented in the header block
- Defensive DSR normalization handles both `['ALL']` and `["ALL"]` format variants
- Disable-users procedure cleanly eliminates unnecessary `ALTER USER` calls for already-disabled users

Leaving a few inline suggestions, one of which is worth addressing before merge.

admin/users.sql
```diff
@@ -162,32 +162,148 @@ CREATE USER IF NOT EXISTS GENIE_SERVICE
...
+                BREAK;
+            END IF;
+        END LOOP;
+        CLOSE role_cursor;
+
+        -- Only non-service users without excluded roles are treated as analysts.
+        should_enable_all := (NOT is_service_user) AND (NOT is_excluded_by_role);
+
+        IF (should_enable_all) THEN
+            IF (normalized_dsr_value NOT IN ('[\'ALL\']', '["ALL"]')) THEN
```
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

🔴 **Bug:** `'[\'ALL\']'` uses backslash escaping to embed a single quote, but this only works when `ESCAPE_STRING_LITERALS = TRUE`. That setting defaults to `FALSE` on newer Snowflake accounts — on those, the `'` ends the string early, producing a syntax error or silently wrong comparison logic.

The portable SQL approach is to double the single quote:

Solution Suggestion:
1. Replace `'[\'ALL\']'` with `'[''ALL'']'`
2. Corrected line: `IF (normalized_dsr_value NOT IN ('[''ALL'']', '["ALL"]')) THEN`

admin/users.sql
```diff
@@ -162,32 +162,148 @@ CREATE USER IF NOT EXISTS GENIE_SERVICE
...
     user_cursor CURSOR FOR 
-        SELECT "name", "default_secondary_roles" 
+        SELECT "name", "type", "default_secondary_roles" 
         FROM TABLE(RESULT_SCAN(LAST_QUERY_ID())) 
-        WHERE "name" <> 'SNOWFLAKE'; -- A cursor over our users
+                WHERE "name" <> 'SNOWFLAKE'
+                        AND LOWER("name") NOT IN (
+                            'joe.smith@sagebase.org',
+                            'joni.harker@sagebase.org'
+                    ); -- Jumpcloud-managed users
```
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

🟡 **Nit:** The `WHERE` and `AND` clauses are over-indented relative to the `FROM` above them. The previous cursor used consistent 8-space alignment. Suggested formatting:

```sql
        FROM TABLE(RESULT_SCAN(LAST_QUERY_ID()))
        WHERE "name" <> 'SNOWFLAKE'
            AND LOWER("name") NOT IN (
                'joe.smith@sagebase.org',
                'joni.harker@sagebase.org'
            ); -- Jumpcloud-managed users
```

admin/users.sql
```diff
@@ -162,32 +162,148 @@ CREATE USER IF NOT EXISTS GENIE_SERVICE
...
+        -- Normalize current value so comparisons work across formatting variants.
+        normalized_dsr_value := UPPER(COALESCE(dsr_value, ''));
+        -- Service users are excluded by explicit type and by service-like username.
+        is_service_user := (UPPER(COALESCE(user_type, '')) = 'SERVICE')
+            OR (POSITION('service' IN LOWER(username)) > 0);
+        is_excluded_by_role := FALSE;
+
+        -- Inspect role grants for this user to detect developer/admin exclusions.
+        EXECUTE IMMEDIATE
+            'SHOW GRANTS TO USER IDENTIFIER(''' || :quoted_username || ''')';
```
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

🟡 **Nit:** Once `is_service_user = TRUE`, `should_enable_all` will be `FALSE` regardless of `is_excluded_by_role`, so this `SHOW GRANTS` call is wasted work for every service account.

Solution Suggestion:
1. Wrap the `EXECUTE IMMEDIATE` + role cursor block in `IF (NOT is_service_user) THEN ... END IF;`
2. The `should_enable_all` logic is unchanged — it already short-circuits correctly

admin/users.sql
```diff
@@ -162,32 +162,148 @@ CREATE USER IF NOT EXISTS GENIE_SERVICE
...
-        WHERE "name" <> 'SNOWFLAKE'; -- A cursor over our users
+                WHERE "name" <> 'SNOWFLAKE'
+                        AND LOWER("name") NOT IN (
+                            'joe.smith@sagebase.org',
+                            'joni.harker@sagebase.org'
+                    ); -- Jumpcloud-managed users
+    role_cursor CURSOR FOR
+        SELECT "name"
+        FROM TABLE(RESULT_SCAN(LAST_QUERY_ID()))
+        WHERE "granted_on" = 'ROLE'; -- A cursor over roles granted to a user
```
jaymedina, line 196: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

🔵 **FYI:** `role_cursor` is declared at the top of the block but relies on `LAST_QUERY_ID()` pointing to a `SHOW GRANTS` result when it's opened. This works correctly as written (it's always opened immediately after the `EXECUTE IMMEDIATE`), but the coupling is subtle and fragile to reordering. A short comment like `-- opened after each SHOW GRANTS call` would help future maintainers.

PR #324, opened 2026-05-14
Update dbt environment configuration instructions

philerooski:
# Problem

It's unclear what the best way to configure the database and schema for our dbt model deployment.

# Solution

I standardized how we configure the model database/schema by consolidating schema configuration in the dbt project file for non-mart models, establishing the model properties file as the appropriate place to configure the schema for mart models, and providing instructions on how to configure the database appropriately based on the deployment environment.

# Testing

 
I created a test database and deployed models to it according to the new guidance.

### DIFF
--- a/transform/README.md
+++ b/transform/README.md
@@ -1,8 +1,8 @@
-# DBT
+# dbt
 
-A dbt project to transform source data ingested into Snowflake into reusable resources.
+A dbt project to transform source data in Snowflake into reusable resources.
 
-dbt works with data that has already been loaded into Snowflake. We can specify transformations of data (see "Models") in a way which align with dbt's source -> staging -> intermediate -> mart paradigm. See the [dbt documentation on project structure](https://docs.getdbt.com/best-practices/how-we-structure/1-guide-overview) for an overview.
+dbt helps us manage downstream data models derived from existing data in Snowflake. One of the more powerful abstractions of dbt is that it enables us to independently manage how data is modeled from how models are materialized in the data warehouse. To _model our data_, we specify transformations in a way which aligns with dbt's [source -> staging -> intermediate -> mart paradigm](https://docs.getdbt.com/best-practices/how-we-structure/1-guide-overview). How we _materialize our data models_ as particular object types (tables/views/etc.) in specific databases/schemas is configured separately.
 
 # Use-Cases
 
@@ -36,7 +36,7 @@ transform:
       threads: 1
 ```
 
-# Models
+# Data Models
 
 Models are categorized according to their function in the dbt model paradigm.
 
@@ -52,7 +52,7 @@ Source data is already loaded into Snowflake tables and dbt is not involved in t
 [Intermediate models](https://docs.getdbt.com/best-practices/how-we-structure/3-intermediate) are derived from one or more staging or intermediate models, typically by joining upon their keys. This is the most flexible layer of models, and serves as an intermediary layer between the staging and mart models.
 
 > [!NOTE]
-> DBT recommends organizing intermediate models by business domain, rather than with respect to the data's source system. We stray from this recommendation by consolidating our staging models into a "universal" intermediate model that can be used by multiple business domains. For example, [`int_synapse_acl.sql`](./models/intermediate/synapse/int_synapse_acl.sql) joins ACL-related tables from the staging models to create a single, unified ACL interface -- rather than having multiple business domains duplicate this logic. 
+> dbt recommends organizing intermediate models by business domain, rather than with respect to the data's source system. We stray from this recommendation by consolidating our staging models into a "universal" intermediate model that can be used by multiple business domains. For example, [`int_synapse_acl.sql`](./models/intermediate/synapse/int_synapse_acl.sql) joins ACL-related tables from the staging models to create a single, unified ACL interface -- rather than having multiple business domains duplicate this logic. 
 
 ## Mart Models
 
@@ -62,11 +62,15 @@ Mart models are organized into subdirectories based on their target database:
 - **`marts/synapse_data_warehouse/`**: Models deployed to the `synapse_data_warehouse` database (or corresponding dev database)
 - **`marts/sage/`**: Models deployed to the `sage` database (production only)
 
-# Deployment Environments
+# Model Materialization
 
-This project deploys its models to different environments, each serving a different purpose:
+As mentioned earlier, how a data model is materialized is configured independently of the data model itself. The following sections primarily discuss _where_ we materialize models – since that's most pertinent when deploying models to a development database – although how a model is materialized, whether as a table, view, or something else, is configured in a similar way. See [model properties](https://docs.getdbt.com/reference/model-configs) for all the configuration settings we can make upon data models.
 
-## Synapse Data Warehouse
+## Deployment Environments
+
+This project deploys its models to different environments, each serving a different subset of data models:
+
+### Synapse Data Warehouse
 
 **Database**: `SYNAPSE_DATA_WAREHOUSE` (production) or `SYNAPSE_DATA_WAREHOUSE_DEV` (staging) or `SYNAPSE_DATA_WAREHOUSE_DEV_{my_branch}` (development).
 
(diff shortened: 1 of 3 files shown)


transform/README.md
```diff
@@ -67,52 +86,95 @@ Analyst-friendly models that may depend on Synapse data warehouse models. Models
...
+models:
+  transform:
+    marts:
+      sage:
+        +database: SYNAPSE_DATA_WAREHOUSE_DEV_MY_FEATURE 
+```
+
+To change the destination schema, edit the model properties file for the relevant model (for example, [`models/marts/sage/_sage__models.yml`](./models/marts/sage/_sage__models.yml)) and update the model's `config.schema` value.
+
+For example, a model can be redirected to a schema which actually exists in your development database:
```
linglp: Should we tell people to deploy in `rds_raw` instead? Since when I deployed to `synapse`, I got the error: 
```
[0m18:59:45  Snowflake adapter: Snowflake query id: 01c43c93-0005-65a9-0004-7c7a04e47452
[0m18:59:45  Snowflake adapter: Snowflake error: 002742 (01000): SQL compilation error: Failed to refresh dynamic table with refresh_trigger INITIAL at data_timestamp 1778266785182 because of the error: SQL compilation error: Target table failed to refresh: SQL compilation error:
Failure during expansion of view 'DATA_ACCESS_SUBMISSION_DASHBOARD': SQL compilation error:
Failure during expansion of view 'INT_SYNAPSE_ACCESS_REQUIREMENT': SQL compilation error:
Failure during expansion of view 'STG_SYNAPSE__ACCESS_REQUIREMENT': SQL compilation error:
Database 'SYNAPSE_DATA_WAREHOUSE_DEV' does not exist or not authorized. (Note: the primary role is the owner role of the dynamic table)
```

linglp: If we deploy to `synapse`, can we add an explanation of future grants?

transform/README.md
```diff
@@ -67,52 +86,95 @@ Analyst-friendly models that may depend on Synapse data warehouse models. Models
...
+For example:
+```
+models:
+  transform:
+    marts:
+      sage:
+        +database: SYNAPSE_DATA_WAREHOUSE_DEV_MY_FEATURE 
+```
+
+To change the destination schema, edit the model properties file for the relevant model (for example, [`models/marts/sage/_sage__models.yml`](./models/marts/sage/_sage__models.yml)) and update the model's `config.schema` value.
```
linglp: I think even after I updated `./models/marts/sage/_sage__models.yml`, I still have to update `.dbt/profiles.yml`. Otherwise I will get: 
```
15:05:30  Finished running 1 dynamic table model in 0 hours 0 minutes and 5.64 seconds (5.64s).
15:05:30  
15:05:30  Completed with 1 error, 0 partial successes, and 0 warnings:
15:05:30  
15:05:30  Failure in model data_access_submission_dashboard (models/marts/sage/data_access_submission_dashboard.sql)
15:05:30    Database Error in model data_access_submission_dashboard (models/marts/sage/data_access_submission_dashboard.sql)
  002003 (02000): SQL compilation error:
  Schema 'SYNAPSE_DATA_WAREHOUSE_DEV_TEST.DUMMY' does not exist or not authorized.

```

linglp: oh never mind. I realized that this happened because I did: 
```
dbt run --target dev --select data_access_submission_dashboard
```
instead of:
```
dbt run -s +data_access_submission_dashboard   
```

transform/README.md
```diff
@@ -67,52 +86,95 @@ Analyst-friendly models that may depend on Synapse data warehouse models. Models
...
+
 ## Deploy Specific Models
 
-We can specify a specific model to be deployed as well. dbt will handle the model dependencies:
+We can specify a specific model to be deployed, as well. dbt will handle the model dependencies:
 ```bash
+# Deploy just this model
 dbt run --select stg_synapse__data_access_submission_status
+
+# Deploy this model and all upstream models
```
linglp: Since data_access_submission_dashboard.sql references upstream models via `{{ ref('int_synapse_data_access_submission_enriched') }}`, and generate_schema_name only controls where a model is created (not where refs resolve to), the ref compiles to `SYNAPSE_DATA_WAREHOUSE_DEV_TEST.dummy.int_synapse_data_access_submission_enriched` — the schema coming from that upstream model's own resolved schema (target.schema = dummy). The fix is to build all upstream models first using `dbt run -s +data_access_submission_dashboard`, so that the upstream model exists in the dev database before the dashboard model tries to read from it.

My suggestion is that here we should add that `--select` alone **does NOT build** upstream dependencies.

And we should also add a section about `SAGE`: 
```
Sage mart models reference upstream intermediate models that resolve to your `target.schema` in `~/.dbt/profiles.yml`. When developing locally, build upstream dependencies first with `dbt run --select +<model-name>`, or ensure the intermediate models already exist in your dev schema.
```

transform/README.md
```diff
@@ -67,52 +86,95 @@ Analyst-friendly models that may depend on Synapse data warehouse models. Models
 > [!IMPORTANT]
 > Sage models are configured to deploy **only in production** (`target.name == 'prod'`). They will be automatically skipped in development environments since corresponding test databases may not exist.
 
-# Running DBT
+# Deploying dbt models
 
-DBT models can be deployed to the databases specified in [`dbt_project.yml`](./dbt_project.yml), and will default to being deployed to the database specified in your DBT profile (`~/.dbt/profiles.yml`). This can be particularly useful when deploying to a development environment, since we don't need to modify any configuration in the DBT project itself.
+There are multiple ways to specify where dbt models are deployed. The following list can be read as a hierarchy, where database/schema configurations in later items supersede any configuration set in preceding items:
```
linglp: How about adding a note here: 

This hierarchy controls where a model is **CREATED**. It does not affect how `{{ ref() }}` resolves upstream dependencies — refs always resolve to wherever the upstream model is configured to be created, which for most models defaults to `target.schema` in `profiles.yml`

Otherwise I think it is confusing because where a model is created is different from where`ref()` resolves to

transform/README.md
philerooski: That error suggests you tried to deploy to `SYNAPSE_DATA_WAREHOUSE_DEV`, which you don't have write privileges to. It doesn't have anything to do with the schema.

transform/README.md
philerooski: I think we could establish this a little higher up, actually. Developers should already understand how dbt abstracts the data model from its manifestation before they get into the details around how we put this into practice. `ref` is an aspect of the data model, whereas the database/schema config is a manifestation detail.

transform/README.md
philerooski: Thanks for testing this.

> The fix is to build all upstream models first using dbt run -s +data_access_submission_dashboard, so that the upstream model exists in the dev database before the dashboard model tries to read from it.

If I understand you correctly, when you tried to deploy the `data_access_submission_dashboard` model on its own without first deploying its upstream dependencies, you saw an error about the `SYNAPSE_DATA_WAREHOUSE_DEV_TEST.dummy.int_synapse_data_access_submission_enriched` view not existing? If so, that's a peculiar edge case: why didn't `ref` respect the schema setting for intermediate synapse models in `dbt_project.yml`?:

```
models:
  transform:
    ...
    intermediate:
      synapse:
        +schema: rds_raw
```

In any case, I can clarify the behavior of `select`.

philerooski: I also found another thing that could use some improvement. Presently, I configure `sage` models to deploy to the `SAGE` database only when `target.name = prod`. This is overly restrictive, because we sometimes want to deploy these models to our development database using a non-prod target. I think its safe enough to make the target constraint a convention, rather than something built-in to the dbt project file.

But I think it makes sense to keep the `+database: sage` line in the project file, despite the fact that we sometimes want to deploy these models to our development database. It does require the developer to understand that they need to go into the dbt project file and either comment out that section or replace the database name (once again) with their development database name. But if we only enforce deploying non-synapse-data-warehouse models to non-synapse-data-warehouse databases by convention (as part of the CI), then I think we give up too much of the utility that we get from using the dbt project file to configure where our models are deployed.

transform/models/staging/synapse/_synapse__sources.yml
```diff
@@ -1,6 +1,5 @@
 sources:
   - name: rds
-    database: "{{ 'SYNAPSE_DATA_WAREHOUSE' if target.name == 'prod' else 'SYNAPSE_DATA_WAREHOUSE_DEV' }}"
```
linglp, line 3: How about doing: 
```
database: "{{ target.database }}"
```
so that you use the target database instead of hard coded database name?

transform/README.md
linglp: Okay thanks for explaining @philerooski ! I realized that I missed the changes in `dbt_project.yml` in my test branch. After adding those, everything seems to be working as expected.

philerooski (commented):
placeholder

transform/models/staging/synapse/_synapse__sources.yml
philerooski: Isn't that already the behavior if we omit `database` here?

linglp: Yes, I understand now after https://sagebionetworks.slack.com/archives/C02M7QR8D46/p1778878930429669

linglp (approved):
LGTM! Thanks for your hard work

transform/README.md
```diff
@@ -1,8 +1,8 @@
-# DBT
+# dbt
 
-A dbt project to transform source data ingested into Snowflake into reusable resources.
+A dbt project to transform source data in Snowflake into reusable resources.
 
-dbt works with data that has already been loaded into Snowflake. We can specify transformations of data (see "Models") in a way which align with dbt's source -> staging -> intermediate -> mart paradigm. See the [dbt documentation on project structure](https://docs.getdbt.com/best-practices/how-we-structure/1-guide-overview) for an overview.
+dbt helps us manage downstream data models derived from existing data in Snowflake. One of the more powerful abstractions of dbt is that it enables us to independently manage how data is modeled from how models are materialized in the data warehouse. To _model our data_, we specify transformations in a way which aligns with dbt's [source -> staging -> intermediate -> mart paradigm](https://docs.getdbt.com/best-practices/how-we-structure/1-guide-overview). How we _materialize our data models_ as particular object types (tables/views/etc.) in specific databases/schemas is configured separately.
```
jaymedina, line 5: nit: _How **and where** we _materialize our data models_ as particular object types (tables/views/etc.) in specific databases/schemas is configured separately._

transform/README.md
```diff
@@ -12,17 +12,36 @@ We currently only use dbt with Synapse RDS snapshot data.
...
 
-After installing, configure your `~/.dbt/profiles.yml` by following the instructions in the previous link. This is where you can specify both your authentication and deployment environment (i.e., database/schema) information. dbt models can be deployed to any schema which your role has write access to.
+After installing, configure your `~/.dbt/profiles.yml` by following the instructions in the previous link. This is where you can specify both your authentication and deployment environment (i.e., database/schema) information.
 
 > [!IMPORTANT]
 > The profile name in `~/.dbt/profiles.yml` must match the `profile` field in `dbt_project.yml`. This project uses `profile: 'transform'`, so your `profiles.yml` must have a top-level key named `transform`.
+>
+> **Additionally**, you need to specify the `database` you will deploy dbt models to in this file. A `schema` must also be set, but we override this value in more specific configurations, so it can be any value.
```
jaymedina: nit: can you add a blurb, and modify the example snippet, showing how you can switch between different databases based on what `target` you choose?

transform/dbt_project.yml
```diff
@@ -25,11 +25,16 @@ models:
       relation: true
       columns: true
     +contract: {enforced: true}
+    staging:
+      synapse:
+        +schema: rds_raw
+    intermediate:
+      synapse:
+        +schema: rds_raw
```
jaymedina, line 33: does this mean `dbt` is going to try to put staging/int models into an `rds_raw` schema in `sage` if a user configures that are their database in their `profiles.yml`?

transform/README.md
```diff
@@ -56,67 +79,94 @@ The majority of models deploy here, including:
...
 **Database**: `SAGE` (production only)
 
 Analyst-friendly models that may depend on Synapse data warehouse models. Models in [`models/marts/sage/`](./models/marts/sage/) deploy here.
 
-> [!IMPORTANT]
-> Sage models are configured to deploy **only in production** (`target.name == 'prod'`). They will be automatically skipped in development environments since corresponding test databases may not exist.
+## Deploying dbt models
 
-# Running DBT
+There are multiple ways to configure a data model, and more specify how to configure where these models are deployed. The following list can be read as a hierarchy, where database/schema configurations in later items supersede any configuration set in preceding items:
```
jaymedina: typo: and more **specifically**

transform/README.md
```diff
@@ -56,67 +79,94 @@ The majority of models deploy here, including:
...
+```yaml
+models:
+  - name: data_access_submission_dashboard
+    config:
+      schema: synapse
+```
+
+# Developer Guidelines
 
-### DBT Resources:
```
jaymedina, line 117: I would keep a **Resources** section that consolidates all the references mentioned throughout the README.md and adds a few extra. Would be helpful!

jaymedina (approved):
LGTM! thanks for updating the docs

transform/dbt_project.yml
philerooski: Yes, if you set your target's database to `SAGE`, then models which aren't configured with a more specific database will be deployed there.

transform/README.md
philerooski: I added a TLDR section for easy reference. I can't vouch for how useful it will be to link to generic dbt resources so I tried to keep things more focused this way.

PR #326, opened 2026-05-19
[SNOW-479] Bootstrap *HIGH PRIORITY* `RDS_LANDING` tables

jaymedina:
# Problem
Ticket: https://sagebionetworks.jira.com/browse/SNOW-479

We move forward in the pipeline implementation plan by creating an initial set of `RDS_LANDING` tables. We prioritize the ones currently being used downstream in our INT models: https://github.com/Sage-Bionetworks/snowflake/pull/278 (+ `PRINCIPAL_ALIAS`). As such, the data types we are making tables for are:

- access_approval
- acl
- acl_resource_access
- acl_resource_access_type
- data_access_submission
- data_access_submission_accessor_change
- data_access_submission_status
- data_access_submission_submitter
- data_access_request
- access_requirement
- access_requirement_project
- access_requirement_revision
- data_access_notification
- principal_alias

---

# Solution

- [x] Introduce a `FILE FORMAT` object for parquet file types so we can infer the schemas of the S3 parquet files to build our tables (`V2.71.0__create_parquet_file_format.sql`)
- [x] Introduce the initial set of tables needed for our current INT models (`V2.71.1__create_high_priority_landing_tables.sql`)

**Context on `V2.71.1__create_high_priority_landing_tables.sql` complexity:**

**[SNOW-475](https://sagebionetworks.jira.com/browse/SNOW-475)** is blocked pending Platform's decision on the S3 path structure for dev/prod parquet objects.

**[SNOW-479](https://sagebionetworks.jira.com/browse/SNOW-479)** can still be done, but is impacted by 475 being blocked. The impact is it adds complexity and flimsiness to the schemas of the initial set of tables:

The `test_with_clone` job in this PR was initially failing because parquet file paths are inconsistent across DPE's RDS snapshots dev/prod buckets, meaning `{{database_name}}` isn't the only dynamic value in the `LOCATION` parameter of the `V2.71.1` script. I have to work around SNOW-475's blockage by making the entire path dynamic based on deployment environment. Two known limitations of the current approach:

1. I'm targeting a single specific stack to generate `LOCATION` values for both environments, rather than sampling across equivalent parquet files for a given data type.
2. Related, to that: The schema inference file is effectively hardcoded, so if `LOCATION` changes and this script gets rerun for whatever reason, it could break.

That said, I don't think these are worth blocking on. The goal right now is to get an initial set of tables in-place so we can build out the task graph and nail down the observability strategy. If tables come out malformed, that's a useful test of the observability layer anyway, and any path changes can be patched later.

---

# Testing
See the corresponding database clone of this feature branch on Snowsight.

[SNOW-475]: https://sagebionetworks.jira.com/browse/SNOW-475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-479]: https://sagebionetworks.jira.com/browse/SNOW-479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/synapse_data_warehouse/rds_landing/V2.71.0__create_parquet_file_format.sql
+++ b/synapse_data_warehouse/rds_landing/V2.71.0__create_parquet_file_format.sql
@@ -1,0 +1,4 @@
+USE SCHEMA {{database_name}}.RDS_LANDING; --noqa: JJ01,PRS,TMP
+
+CREATE OR REPLACE FILE FORMAT parquet_ff
+    TYPE = PARQUET;
--- a/synapse_data_warehouse/rds_landing/V2.71.0__create_parquet_file_format.sql
+++ b/synapse_data_warehouse/rds_landing/V2.71.0__create_parquet_file_format.sql
@@ -1,4 +1,0 @@
-USE SCHEMA {{database_name}}.RDS_LANDING; --noqa: JJ01,PRS,TMP
-
-CREATE OR REPLACE FILE FORMAT parquet_ff
-    TYPE = PARQUET;
--- a/synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
+++ b/synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
@@ -127,7 +127,7 @@ CREATE TABLE IF NOT EXISTS data_access_notification (
     requirement_id     BIGINT  COMMENT 'The access requirement identifier',
     recipient_id       BIGINT  COMMENT 'User ID of the notification recipient',
     access_approval_id BIGINT  COMMENT 'The identifier of the associated access approval',
-    sent_on            BIGINT  COMMENT 'Epoch milliseconds when the notification was sent',
+    sent_on            TIMESTAMP_NTZ(9)  COMMENT 'Timestamp when the notification was sent',
     message_id         BIGINT  COMMENT 'Unique identifier of the message',
     etag               VARCHAR COMMENT 'Entity tag for optimistic concurrency control (36-character UUID)'
 );
--- a/synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
+++ b/synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
@@ -7,7 +7,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.ACCESS_APPROVAL/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -19,7 +19,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.ACL/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -31,7 +31,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.ACL_RESOURCE_ACCESS/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -43,7 +43,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.ACL_RESOURCE_ACCESS_TYPE/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -55,7 +55,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.DATA_ACCESS_SUBMISSION/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -67,7 +67,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.DATA_ACCESS_SUBMISSION_ACCESSOR_CHANGE/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -79,7 +79,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.DATA_ACCESS_SUBMISSION_STATUS/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -91,7 +91,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.DATA_ACCESS_SUBMISSION_SUBMITTER/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -103,7 +103,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.DATA_ACCESS_REQUEST/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -115,7 +115,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.ACCESS_REQUIREMENT/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -127,7 +127,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.ACCESS_REQUIREMENT_PROJECT/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -139,7 +139,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.ACCESS_REQUIREMENT_REVISION/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
@@ -151,7 +151,7 @@ USING TEMPLATE (
     FROM TABLE(
         INFER_SCHEMA(
             LOCATION => '@{{database_name}}.RDS_LANDING.RDS_SNAPSHOTS_STAGE/rds-snapshot/prod-589-db-2026-05-19/prod589.DATA_ACCESS_NOTIFICATION/1/',
-            FILE_FORMAT => '{{database_name}}.RDS_LANDING.<file_format_name>'
+            FILE_FORMAT => '{{database_name}}.RDS_LANDING.parquet_ff'
         )
     )
 );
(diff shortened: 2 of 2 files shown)


philerooski (commented):
The procedure is slick, but it obscures what we are actually deploying to production. Inferring the schemas was fine for a manual bootstrap solution, but there isn't a need to be as hand-wavy with a long-term production solution.

If the vision is to maintain these schemas in a single place (the dbt source file), I would recommend starting there. Not as something which we write SQL against to do automated DDL generation during deployment (which is sort of what's happening here, with INFER_SCHEMA), but as something you can point an agent to so that it can write individual schemachange scripts for each table, containing vanilla DDL:
```
CREATE OR REPLACE TABLE acl (
  id NUMBER UNIQUE NOT NULL,
  owner_id NUMBER NOT NULL,
  owner_type VARCHAR,
  ...
)
```

This way it's explicit that the schema we create conforms to what dbt expect, and isn't dependent on the format of the data provided to us and the black box behavior of INFER_SCHEMA.

I would also verify that we aren't missing any constraints in the dbt source file. Everything in the MySQL DDL ought to be present in the dbt constraints, as well.

jaymedina:
@philerooski you're right, I can just refer to the existing STG models! I'll do it this way. Thanks

synapse_data_warehouse/rds_landing/V2.71.0__create_parquet_file_format.sql
thomasyu888, line 1: Is this still needed?

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
```diff
@@ -0,0 +1,143 @@
+USE SCHEMA {{database_name}}.RDS_LANDING; --noqa: JJ01,PRS,TMP
+
+-- access_approval
+CREATE TABLE IF NOT EXISTS lan_synapse_access_approval (
```
thomasyu888: Is there value in adding this information here, instead of DBT (or both)

```
  PRIMARY KEY (`REQUIREMENT_ID`, `ACCESSOR_ID`, `REQUIREMENT_VERSION`, `SUBMITTER_ID`),
  UNIQUE KEY (`ID`),
  CONSTRAINT `ACCESS_APPROVAL_REQUIREMENT_ID_FK` FOREIGN KEY (`REQUIREMENT_ID`) REFERENCES `ACCESS_REQUIREMENT` (`ID`) ON DELETE CASCADE,
  CONSTRAINT `ACCESS_APPROVAL_CREATED_BY_FK` FOREIGN KEY (`CREATED_BY`) REFERENCES `USER_GROUP` (`ID`) ON DELETE CASCADE,
  CONSTRAINT `ACCESS_APPROVAL_MODIFIED_BY_FK` FOREIGN KEY (`MODIFIED_BY`) REFERENCES `USER_GROUP` (`ID`) ON DELETE CASCADE,
  CONSTRAINT `ACCESS_APPROVAL_SUBMITTER_ID_FK` FOREIGN KEY (`SUBMITTER_ID`) REFERENCES `USER_GROUP` (`ID`) ON DELETE CASCADE,
  CONSTRAINT `ACCESS_APPROVAL_ACCESSOR_ID_FK` FOREIGN KEY (`ACCESSOR_ID`) REFERENCES `USER_GROUP` (`ID`) ON DELETE CASCADE,
  INDEX `ACCESS_APPROVAL_EXPIRED_ON_INDEX` (`EXPIRED_ON`)
```

Discuss in more detail if this should be included both in SQL and DBT YAML

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
```diff
@@ -0,0 +1,143 @@
...
+    id                  BIGINT    COMMENT 'Unique identifier for the access approval record',
+    requirement_id      BIGINT    COMMENT 'Which requirement this approval satisfies',
+    requirement_version BIGINT    COMMENT 'Version number of the access requirement at the time of approval',
+    created_by          BIGINT    COMMENT 'Identifier of the user who created this access approval',
+    created_on          BIGINT    COMMENT 'Epoch milliseconds when the access approval was created',
+    modified_by         BIGINT    COMMENT 'Identifier of the user who last modified this access approval',
+    modified_on         BIGINT    COMMENT 'Epoch milliseconds when the access approval was last modified',
```
thomasyu888, line 11: will the conversion to NTZ time happen in the staging layer? (yes)

RDS_LAND -> RDS_RAW w/ STAGE (views) -> INT -> MART?

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
```diff
@@ -0,0 +1,143 @@
...
+
+-- data_access_submission_status
+CREATE TABLE IF NOT EXISTS lan_synapse_data_access_submission_status (
+    submission_id BIGINT  COMMENT 'The data access submission identifier',
+    created_by    BIGINT  COMMENT 'Identifier of the user who created this status record',
+    created_on    BIGINT  COMMENT 'Epoch milliseconds when the status was created',
+    modified_by   BIGINT  COMMENT 'User ID of the person who last modified this status',
+    modified_on   BIGINT  COMMENT 'Epoch milliseconds when the submission status was last modified',
+    state         VARCHAR COMMENT 'Current state. One of: Submitted, Approved, Rejected, Cancelled',
+    reason        BINARY  COMMENT 'Binary blob containing the reason for the current state. Null for records before 2019 due to invalid UTF-8.'
```
thomasyu888, line 70: when are things BINARY vs VARCHAR?

thomasyu888 (commented):
I'll defer to @philerooski for a final pass here.

As a first pass, I did a comparison per DDL in this PR against here https://github.com/Sage-Bionetworks/Synapse-Repository-Services/tree/develop/lib/jdomodels/src/main/resources/schema to look at any differences

philerooski:
Were these table/column comments generated by AI?

philerooski:
Slightly more than a nit: Let's not prefix the table names with `lan_`. This is a dbt convention to easily distinguish / grep for models situated in a particular data layer. But these tables aren't part of the dbt framework, and not even dbt recommends prefixing the source tables -- tables in the RDS_RAW schema, in our case, which are just downstream of these tables -- with any particular prefix.

imo we should keep the same name as the MySQL table counterpart, since these are effectively Snowflake clones of those tables.

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
philerooski: The [MySQL DDL](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/develop/lib/jdomodels/src/main/resources/schema/DataAccessSubmissionStatus-ddl.sql) for this table represents the `reason` column as a [`blob`](https://dev.mysql.com/doc/refman/8.0/en/blob.html) data type, which can be any set of bits, including something that could be a text encoding (VARCHAR). In this case, it really is a UTF-8 encoding, but rather than trying to encode the bits at load time, we encode in the staging layer: https://github.com/Sage-Bionetworks/snowflake/blob/34078d4395d34e23ba7760a381abf3b626e674e1/transform/models/staging/synapse/stg_synapse__data_access_submission_status.sql#L19-L22

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
jaymedina: @philerooski do you remember what we agreed on during our discussion of this at Code Spaces last Friday? I don't see this convo captured in the [meeting notes](https://docs.google.com/document/d/1RkmSaMXkbNLRsPgLaDS9J9fD4twAGxRjSCIb8ua9Vlw/edit?usp=sharing)

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
jaymedina: @philerooski could you remind me how you plan to distinguish between the `RAW` and the `STG` models in the `RDS_RAW` schema?

jaymedina:
@philerooski:

> _Were these table/column comments generated by AI?_

Yes, I used an agent to reference the column descriptions of the staging models based on the matching or **analogous*** column names.

**analogous*** = For example, my agent was able to map `request_serialized` to its renamed counterpart `data_access_request_raw` in the `data_access_request` staging model, and pull the comment and data type that way.

Important caveat: The agent was also able to infer the previous state of metadata based on the transformation you've applied to it at the staging layer. For example, even though `state_reason` is a `VARCHAR` data type, the agent inferred that its transformation in the `stg_synapse__data_access_submission_status.sql` implies it was a binary blob before.

> _imo we should keep the same name as the MySQL table counterpart, since these are effectively Snowflake clones of those tables._

That makes sense to me. Updated!

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
jaymedina: > _when are things BINARY vs VARCHAR?_

The **Important caveat:** note in my comment [here](https://github.com/Sage-Bionetworks/snowflake/pull/326#issuecomment-4565665270) explains how I used my agent to infer the data types of these columns, based on how (if at all) the columns were transformed at the staging layer.

With that said, I went in manually checked the `VARCHAR` and `BINARY`-type columns to ensure they match up with the platform team's MySQL DDL, and I'll run my agent to cross-reference that repository to verify the data types generated here.

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
philerooski: @jaymedina 

The data layering that dbt recommends is:

source -> staging -> intermediate -> mart

dbt doesn't model the source tables, it just assumes that they already exist in the data warehouse. Our source tables are _tables_ in `RDS_RAW`, and the naming convention we use now is to name them the same as their corresponding RDS table, e.g., `ACCESS_APPROVAL`, `ACCESS_REQUIREMENT`, etc.

dbt does model the staging models, and we currently materialize those as views in RDS_RAW. The naming convention we use now is to prefix the corresponding source table name with `stg_`, e.g., `stg_access_approval`, `stg_access_requirement`, etc.

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
philerooski: I think we want to add it to both. There will likely be a lot of duplicated constraints, but in the case where we do actually transform the data in some way (for example, we convert the `reason` column of the `data_access_submission_status` data type from binary to varchar) it's nice to have constraints baked in at every dbt data layer, rather than have ad-hoc constraints on specific columns depending on whether we do any meaningful transform upon them.

philerooski (approved):
Whoops I accidentally left my comments in a review. I want to look at the column definitions more critically before we merge but this is looking good overall.

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
jaymedina: Actually, wouldn't we want to assign PKs and FKs to the INT/MART models, since a column type can be considered a PK in one model, but a FK in another? We are mix-and-matching these columns in our models, so they would never be strictly PK/FK/neither.

jaymedina:
> I want to look at the column definitions more critically before we merge but this is looking good overall.

Sounds good, thanks!

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
```diff
@@ -0,0 +1,143 @@
...
+-- access_approval
+CREATE TABLE IF NOT EXISTS access_approval (
+    id                  BIGINT    COMMENT 'Unique identifier for the access approval record',
+    requirement_id      BIGINT    COMMENT 'Which requirement this approval satisfies',
+    requirement_version BIGINT    COMMENT 'Version number of the access requirement at the time of approval',
+    created_by          BIGINT    COMMENT 'Identifier of the user who created this access approval',
+    created_on          BIGINT    COMMENT 'Epoch milliseconds when the access approval was created',
+    modified_by         BIGINT    COMMENT 'Identifier of the user who last modified this access approval',
+    modified_on         BIGINT    COMMENT 'Epoch milliseconds when the access approval was last modified',
+    submitter_id        BIGINT    COMMENT 'User ID of the person who submitted the approval request',
```
philerooski: re `submitter_id`: This is the identifier of the user who created the data access request / data access submission associated with this approval. It corresponds to the `created_by` column in the `data_access_submission` table. The current definition swaps data access request / data access submission with "approval request".

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
```diff
@@ -0,0 +1,143 @@
...
+    id         BIGINT  COMMENT 'Unique identifier for the ACL (Access Control List)',
+    owner_id   BIGINT  COMMENT 'ID of the entity or object that this ACL grants access to',
+    owner_type VARCHAR COMMENT 'Type of entity or object that owns this ACL',
+    created_on BIGINT  COMMENT 'Epoch milliseconds when the ACL was created',
+    etag       VARCHAR COMMENT 'Entity tag for optimistic concurrency control (36-character UUID)'
+);
+
+-- acl_resource_access
+CREATE TABLE IF NOT EXISTS acl_resource_access (
+    id       BIGINT COMMENT 'Unique identifier for this ACL resource access',
```
philerooski: "ACL resource access record", technically

jaymedina (commented):
_Note: This comment has been generated with AI assistance and reviewed by me ✍️_

One type mismatch found against the MySQL DDL — everything else looks good. The landing layer type choices are sensible (ENUMs → `VARCHAR`, `mediumblob` → `BINARY`). See inline comment on `data_access_notification.sent_on`.

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
```diff
@@ -0,0 +1,143 @@
...
+);
+
+-- data_access_notification
+CREATE TABLE IF NOT EXISTS data_access_notification (
+    id                 BIGINT  COMMENT 'Unique identifier for the notification record',
+    notification_type  VARCHAR COMMENT 'Type of notification sent. One of: FIRST_RENEWAL_REMINDER, SECOND_RENEWAL_REMINDER, REVOCATION',
+    requirement_id     BIGINT  COMMENT 'The access requirement identifier',
+    recipient_id       BIGINT  COMMENT 'User ID of the notification recipient',
+    access_approval_id BIGINT  COMMENT 'The identifier of the associated access approval',
+    sent_on            BIGINT  COMMENT 'Epoch milliseconds when the notification was sent',
```
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

🔴 **Bug:** `sent_on` is typed as `BIGINT` with a comment of "Epoch milliseconds", but the [MySQL DDL for `DATA_ACCESS_NOTIFICATION`](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/develop/lib/jdomodels/src/main/resources/schema/DataAccessNotification-ddl.sql) defines this column as `TIMESTAMP(3)` — an actual datetime type, not an integer. Unlike all the other `*_on` columns in this file (which are genuinely stored as epoch-millisecond BIGINTs in MySQL), `SENT_ON` is a real MySQL timestamp. When exported to Parquet via RDS snapshots, it will be written as a Parquet timestamp logical type, not a plain integer. Loading a Parquet timestamp into a Snowflake `BIGINT` column will either silently produce a garbage numeric value or fail at COPY time.

**Solution Suggestion:**
1. Change the type to `TIMESTAMP_NTZ(3)`
2. Update the comment to remove "Epoch milliseconds" (e.g. `'Timestamp when the notification was sent'`)

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
jaymedina: @philerooski let's save this discussion for our code review today, as I still have 1 more PR to deliver before SNOW-479 can be marked as complete, and anything we decide on re: FKs/PKs can be implemented there.

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
jaymedina: I'm using `TIMESTAMP_NTZ(9)` because that is what I'm seeing in the DDL for this table in the `SYNAPSE_RDS_SNAPSHOT` database:

image

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
jaymedina: @philerooski thanks! So if the `RDS_RAW` source tables mirror the `RDS_LANDING` source tables and the `stg_` models are views on top of them, what's the relationship between `RDS_RAW` and `RDS_LANDING`? Are the `RDS_RAW` tables essentially a transformed version of what's in `RDS_LANDING`, and if so wouldn't that transformation logic be worth capturing as dbt models?

jaymedina:
I had an agent review this PR using the MySQL DDL in SRS and addressed remaining issues. The tables are looking good in the clone, so I'm going to merge this. Thanks all for the review and discussion!

image

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
jaymedina: @thomasyu888 

Phil and I have agreed to not implement constraints to the columns in the RDS_LANDING source tables because the constraints would be incomplete: The MySQL source schemas define FK relationships that reference tables we don't collect, so any constraints we added would misrepresent the actual integrity of the data.

But I've made a Jira ticket to revisit this if we want to later: https://sagebionetworks.jira.com/browse/SNOW-489

thomasyu888: I am ok with this, and just so I'm on the same page. Can you give me an example of this scenario?

synapse_data_warehouse/rds_landing/V2.71.1__create_high_priority_landing_tables.sql
philerooski: @jaymedina Let's discuss this on a call sometime.

PR #335, opened 2026-05-21
[SNOW-480] Create initial RDS snapshots task graph for *HIGH PRIORITY* tables

jaymedina:
# Problem

Ticket: [SNOW-480](https://sagebionetworks.jira.com/browse/SNOW-480)

Synapse RDS snapshot data is exported as Parquet files to S3 by the Platform team but there's no recurring pipeline to load it into Snowflake for analytics.

# Solution

Update high priority tables with...
- [x] `stack` column
- [x] `snapshot_date` column 
- [x] `filename` column

Create task graph with...
- [x] root task (to refresh the stage)
- [x] high priority `COPY INTO` tasks
- [x] finalizer task

Implements the Phase 2 daily ingestion pipeline for SNOW-456. A Snowflake Task DAG loads Parquet files from RDS_SNAPSHOTS_STAGE into SYNAPSE_DATA_WAREHOUSE.RDS_LANDING for all 157 record types. The graph fans out through two proxy tasks (working around Snowflake's 100-child-task limit) so all COPY INTO tasks run in parallel. A finalizer task runs unconditionally at the end and posts a Slack notification with load status and row counts regardless of whether the run succeeded or failed.

# Testing

- [x] Test that the task graph of the clone on Snowsight looks as expected, according to the final design.

See clone.

----

- [x] Test that the tasks indeed communicate with and update their corresponding landing tables in `RDS_LANDING`.

See clone.

----

- [x] All high priority tables have been updated after triggering a refresh.

See clone.

----

- [x] Ensure the finalizer task still runs and sends a report even when root task fails
image

----

- [x] Test a partial success scenario
image

5 failed:

image

----

- [x] Test a full success scenario

image

----

- [x] Verify the new columns are in all high priority tables, and populated correctly

I ran this query against all the high priority tables and get expected results:

image

All except for `ACCESS_REQUIREMENT_PROEJCT` which only ingested data from yesterday and today:

image

The reason being that the [2026-06-16 parquet file](https://766808016710-ex2dq7so.us-east-1.console.aws.amazon.com/s3/object/synapse-snowflake-rds-snapshots-dev?region=us-east-1&prefix=rds-snapshot/ACCESS_REQUIREMENT_PROJECT/594/2026-06-16/part-00000-5e946590-f13d-49c0-a328-cb3053eddb69-c000.gz.parquet) is empty:

Screenshot 2026-06-18 at 3 10 41 PM

---- 

- [x] Test that there are no duplicate records after de-duplicating with `snapshot_date`

I ran this test on a few tables to visually check for duplicate `ID`s which would be the PK in this context and found none:

```
SELECT *
FROM access_approval
QUALIFY ROW_NUMBER() OVER (PARTITION BY ID::NUMBER ORDER BY SNAPSHOT_DATE DESC) = 1;
```

Similarly, I ran the following test with `ACCESS_REQUIREMENT_REVISION` and see no difference when deduplicating with `snapshot_date` vs the primary key (in this case `owner_id, number`) which means `snapshot_date` can be used for deduplication!

```
WITH dedup_snapshot_date AS (
    SELECT COUNT(*) AS dedup_snapshot_date_count
    FROM (
        SELECT owner_id
        FROM access_requirement_revision
        QUALIFY ROW_NUMBER() OVER (PARTITION BY owner_id, number ORDER BY SNAPSHOT_DATE DESC) = 1
    )
),
dedup_id AS (
    SELECT COUNT(DISTINCT owner_id, number) AS dedup_id_count
    FROM access_requirement_revision
)
SELECT
    (SELECT COUNT(*) FROM access_requirement_revision) AS total_rows,
    dedup_snapshot_date_count,
    dedup_id_count
FROM dedup_snapshot_date
CROSS JOIN dedup_id;
```

image

[SNOW-480]: https://sagebionetworks.jira.com/browse/SNOW-480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


synapse_data_warehouse/rds_landing/V2.71.3__create_rds_snapshot_tasks.sql
```diff
@@ -0,0 +1,264 @@
...
+
+
+-- ============================================================
+-- Proxy task: no-op fan-out node between the root and the
+-- COPY INTO tasks. Exists to work around Snowflake's 100-child-
+-- task limit per node; add PROXY_TASK_B when child tasks on this
+-- node exceed 100.
+-- ============================================================
+CREATE OR REPLACE TASK PROXY_TASK_A
+    WAREHOUSE = compute_xsmall
```
philerooski: Should we just do serverless for the proxy tasks? It would save us a nominal amount of money:

> Charges for serverless features are calculated based on total usage of snowflake-managed compute resources measured in compute-hours. Compute-Hours are calculated on a per second basis, rounded up to the nearest whole second.

synapse_data_warehouse/rds_landing/V2.71.3__create_rds_snapshot_tasks.sql
```diff
@@ -0,0 +1,264 @@
...
+            $1:MODIFIED_BY::BIGINT         AS modified_by,
+            $1:MODIFIED_ON::BIGINT         AS modified_on,
+            $1:SUBMITTER_ID::BIGINT        AS submitter_id,
+            $1:ACCESSOR_ID::BIGINT         AS accessor_id,
+            $1:EXPIRED_ON::BIGINT          AS expired_on,
+            $1:STATE::VARCHAR              AS state,
+            $1:ETAG::VARCHAR               AS etag
+        FROM @RDS_SNAPSHOTS_STAGE/rds-snapshot/ACCESS_APPROVAL
+    )
+    PATTERN = '.*/[0-9]+/[0-9]{4}-[0-9]{2}-[0-9]{2}/.*\.gz\.parquet';
```
philerooski: Do we need to escape forward slashes? Looks good otherwise.

Image

synapse_data_warehouse/rds_landing/V2.71.2__create_rds_snapshot_tasks.sql
```diff
@@ -0,0 +1,264 @@
...
+            $1:CREATED_BY::BIGINT          AS created_by,
+            $1:CREATED_ON::BIGINT          AS created_on,
+            $1:MODIFIED_BY::BIGINT         AS modified_by,
+            $1:MODIFIED_ON::BIGINT         AS modified_on,
+            $1:SUBMITTER_ID::BIGINT        AS submitter_id,
+            $1:ACCESSOR_ID::BIGINT         AS accessor_id,
+            $1:EXPIRED_ON::BIGINT          AS expired_on,
+            $1:STATE::VARCHAR              AS state,
+            $1:ETAG::VARCHAR               AS etag
+        FROM @RDS_SNAPSHOTS_STAGE/rds-snapshot/ACCESS_APPROVAL
```
philerooski: Should this read `.../rds-snapshot-lambda/...` ?

synapse_data_warehouse/rds_landing/V2.71.2__create_rds_snapshot_tasks.sql
```diff
@@ -0,0 +1,264 @@
...
+            $1:ETAG::VARCHAR          AS etag
+        FROM @RDS_SNAPSHOTS_STAGE/rds-snapshot/PRINCIPAL_ALIAS
+    )
+    PATTERN = '.*/[0-9]+/[0-9]{4}-[0-9]{2}-[0-9]{2}/.*\.gz\.parquet';
+
+-- ============================================================
+-- Resume tasks: children must be resumed before the root task.
+-- Root task is resumed last to activate the schedule.
+-- ============================================================
+ALTER TASK PROXY_TASK_A RESUME;
```
philerooski: Do we need to run this second to last, since this is the parent node of the following tasks?

CONTRIBUTING.md
```diff
@@ -148,6 +148,16 @@ You can go on Snowsight to perform manual inspection of the changes to your sche
...
+> [!IMPORTANT]
+> If your changes include DDL or task behavior that depends on stage metadata refresh (for example, workflows that run `ALTER STAGE ... REFRESH`), verify that DIRECTORY is enabled on the cloned stage before testing.
+>
+> In clone environments, if DIRECTORY is not enabled in your stage, refresh-dependent behavior may fail.
+>
+> Run the following in Snowsight against your cloned database before validation:
+>
+> ALTER STAGE IF EXISTS <your_clone_database>.<your_schema>.<your_stage> SET DIRECTORY = (ENABLE = TRUE);
+> ALTER STAGE IF EXISTS <your_clone_database>.<your_schema>.<your_stage> REFRESH;
+
```
philerooski: This seems hyper-specific to working with tasks and stages, rather than a general contribution guideline. "Perform Inspection using Snowsight" doesn't seem like the right section for this, either.

Although, I'm not really sure where to put this. It's such a niche "gotcha" behavior.

philerooski (changes requested):
I left comments about a few things to check out. I'm not sure if we need any changes, but I'd like to approve one last time just to make sure that we change the schedule of the root task before merge.

synapse_data_warehouse/rds_landing/V2.71.3__create_rds_snapshot_tasks.sql
jaymedina: It looks like no, based on the screenshot, or am I reading this wrong? Also, what tool is this? Looks nifty

synapse_data_warehouse/rds_landing/V2.71.2__create_rds_snapshot_tasks.sql
jaymedina: ooo good catch, thanks!

CONTRIBUTING.md
jaymedina: This doc update is no longer relevant, as I'm going to address the gotcha behavior directly in `test_with_clone.yml`. See [Jira ticket](https://sagebionetworks.jira.com/browse/SNOW-495)

jaymedina: > It's such a niche "gotcha" behavior

tell me about it

synapse_data_warehouse/rds_landing/V2.71.2__create_rds_snapshot_tasks.sql
jaymedina: [Done](https://github.com/Sage-Bionetworks/snowflake/pull/335/commits/250068773107894a95e215ad0ebcd7d9c8faa4cb)! 🫡

synapse_data_warehouse/rds_landing/V2.71.2__create_rds_snapshot_tasks.sql
jaymedina: Done! This was working in my manual tests because I used the correct path in my worksheet, I just hadn't updated it correctly in the SQL script. Thanks again for the catch:

image

synapse_data_warehouse/rds_landing/V2.71.3__create_rds_snapshot_tasks.sql
philerooski: The regex in the screenshot starts out `.*\/...` whereas your regex reads `.*/...`.

The tool is https://rubular.com/ . They had us use this in my intro to CS class in college and I've been using it ever since 😄

jaymedina: Nice! Bookmarked :P

synapse_data_warehouse/rds_landing/V2.71.4__create_rds_snapshot_tasks.sql
```diff
@@ -0,0 +1,275 @@
...
+            $1:CREATED_BY::BIGINT                 AS created_by,
+            $1:CREATED_ON::BIGINT                 AS created_on,
+            $1:ACCESS_REQUIREMENT_VERSION::BIGINT AS access_requirement_version,
+            $1:ETAG::VARCHAR                      AS etag,
+            $1:SUBMISSION_SERIALIZED::BINARY      AS submission_serialized
+        FROM @RDS_SNAPSHOTS_STAGE/rds-snapshot/DATA_ACCESS_SUBMISSION/
+    )
+    -- The parquet files for this data type will contain one or more binary columns
+    -- that cannot be UTF-8 decoded, so BINARY_AS_TEXT must be set to FALSE
+    FILE_FORMAT = (TYPE = PARQUET BINARY_AS_TEXT = FALSE)
```
jaymedina, line 123: When handling Parquet files with one or more BINARY columns containing arbitrary byte data like `...SERIALIZED` (not valid UTF-8), I set `BINARY_AS_TEXT = FALSE` inline in the `COPY INTO` DDL rather than abstracting this logic into a named File Format object.

A separate File Format object for a single config change seems like overkill. Keeping it inline also makes the DDL self-documenting (an engineer can see exactly what's different about this COPY INTO and why it's different, without having to track down a separate object).

synapse_data_warehouse/rds_landing/V2.71.3__add_ingest_metadata_cols_and_clusterby_to_high_priority_landing_tables.sql
```diff
@@ -0,0 +1,113 @@
...
+ALTER TABLE access_approval ADD COLUMN snapshot_date DATE COMMENT 'The date on which the RDS snapshot was taken';
+ALTER TABLE access_approval ADD COLUMN filename VARCHAR COMMENT 'The S3 path of the parquet file from which this record was loaded';
+-- Cluster by snapshot_date
+ALTER TABLE access_approval CLUSTER BY (snapshot_date);
+
+-- ACL
+-- Add ingest metadata columns
+ALTER TABLE acl ADD COLUMN stack INTEGER COMMENT 'The Synapse stack number from which this RDS snapshot was taken';
+ALTER TABLE acl ADD COLUMN snapshot_date DATE COMMENT 'The date on which the RDS snapshot was taken';
+ALTER TABLE acl ADD COLUMN filename VARCHAR COMMENT 'The S3 path of the parquet file from which this record was loaded';
```
jaymedina, line 15: I'm adding the parquet path itself to the `RDS_LANDING` tables for more visibility into the source of the table rows. This makes the `RDS_LANDING` tables also serve as logs to confirm that the `stack` and `snapshot_date` metadata are being derived correctly.

thomasyu888: 🤩 PrOVEnAnCE 🤩

synapse_data_warehouse/rds_landing/V2.71.4__create_rds_snapshot_tasks.sql
```diff
@@ -0,0 +1,317 @@
+USE SCHEMA {{database_name}}.RDS_LANDING; --noqa: JJ01,PRS,TMP
+
+-- ============================================================
+-- Root task: refreshes stage metadata so COPY INTO tasks see
+-- the latest S3 files before any loading begins.
+-- ============================================================
+CREATE OR REPLACE TASK REFRESH_RDS_SNAPSHOTS_STAGE_TASK
+    -- TODO: every minute for testing; change to 3AM ET // 7AM UTC before production rollout.
+    SCHEDULE = 'USING CRON * * * * * America/Los_Angeles'
```
philerooski: Change this.

synapse_data_warehouse/rds_landing/V2.71.4__create_rds_snapshot_tasks.sql
```diff
@@ -0,0 +1,317 @@
...
+ALTER TASK COPY_DATA_ACCESS_SUBMISSION_STATUS_TASK RESUME;
+ALTER TASK COPY_DATA_ACCESS_SUBMISSION_SUBMITTER_TASK RESUME;
+ALTER TASK COPY_DATA_ACCESS_REQUEST_TASK RESUME;
+ALTER TASK COPY_ACCESS_REQUIREMENT_TASK RESUME;
+ALTER TASK COPY_ACCESS_REQUIREMENT_PROJECT_TASK RESUME;
+ALTER TASK COPY_ACCESS_REQUIREMENT_REVISION_TASK RESUME;
+ALTER TASK COPY_DATA_ACCESS_NOTIFICATION_TASK RESUME;
+ALTER TASK COPY_PRINCIPAL_ALIAS_TASK RESUME;
+ALTER TASK PROXY_TASK_A RESUME;
+ALTER TASK REFRESH_RDS_SNAPSHOTS_STAGE_TASK RESUME;
```
philerooski: Can we use `select system$task_dependents_enable('refresh_rds_snapshots_stage_task');` here or no?

synapse_data_warehouse/rds_landing/V2.71.4__create_rds_snapshot_tasks.sql
```diff
@@ -0,0 +1,317 @@
...
+            $1:ACCESSOR_ID::BIGINT         AS accessor_id,
+            $1:EXPIRED_ON::BIGINT          AS expired_on,
+            $1:STATE::VARCHAR              AS state,
+            $1:ETAG::VARCHAR               AS etag,
+            SPLIT_PART(METADATA$FILENAME, '/', 3)::BIGINT AS stack,
+            SPLIT_PART(METADATA$FILENAME, '/', 4)::DATE    AS snapshot_date,
+            METADATA$FILENAME                              AS filename
+        FROM @RDS_SNAPSHOTS_STAGE/rds-snapshot/ACCESS_APPROVAL/
+    )
+    PATTERN = '.*\/[0-9]+\/[0-9]{4}-[0-9]{2}-[0-9]{2}\/.*\.gz\.parquet';
```
philerooski, line 53: We aren't specifying the file format here. I assume it will auto-detect? I was checking out the file format options for Parquet [here](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#type-parquet) and I wasn't sure if some of these options might be useful. For example, the docs actually recommend that we change the default value of USE_VECTORIZED_SCANNER , although how that affects the data we read in is unclear to me. It might be simpler to just omit the FILE_FORMAT entirely, as you've already done here.

philerooski (approved):
The cron schedule needs to be updated, but this is looking good!

synapse_data_warehouse/rds_landing/V2.71.5__create_rds_snapshot_finalizer_task.sql
```diff
@@ -0,0 +1,119 @@
...
+        v_message := '⚠️ No graph status retrieved, or received an unexpected status. DPE team please view task statuses in '
+            || 'snowflake.account_usage.task_history'
+            || ' · *Root Task State*: ' || coalesce(:v_root_task_state, 'NULL')
+            || ' · *Run date*: ' || v_run_date || ' — @team-dpe';
+    end if;
+
+    call system$send_snowflake_notification(
+        snowflake.notification.text_plain(:v_message),
+        -- TODO: point to prod integration when ready
+        snowflake.notification.integration('DEV_SLACK_INGEST_UPDATES')
```
jaymedina, line 109: I'm going to create a separate PR to point these messages to `#slack_ingest_updates` once we're seeing expected behavior on the dev channel.

jaymedina: https://sagebionetworks.jira.com/browse/SNOW-513

PR #337, opened 2026-05-26
[SNOW-433] Add ELITE Usage Metrics Streamlit app

philerooski:
# Problem

Ticket: [SNOW-433](https://sagebionetworks.jira.com/browse/SNOW-433)

# Solution

I converted the dashboard to a Streamlit app and fixed any errors which were either introduced as part of this process or were preexisting.

# Testing

You can test the Streamlit app by either examining it in Snowflake or deploying it locally.

The app has already been deployed to Snowflake, but is only accessible by the schema's respective admin role. If you assume the `SAGE_ADMIN` role or higher, you will be able to view the Streamlit in Snowflake.

You can also deploy the app locally. See `STREAMLIT.md` for instructions.

 

[SNOW-433]: https://sagebionetworks.jira.com/browse/SNOW-433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sage/elite/streamlit/elite_usage_metrics/snowflake.yml
```diff
@@ -0,0 +1,15 @@
+definition_version: 2
+entities:
+  streamlit_app:
+    type: streamlit
+    identifier:
+      name: elite_usage_metrics
+      database: SAGE
+      schema: ELITE
+    query_warehouse: STREAMLIT_XSMALL
```
jaymedina, line 9: I ran into an error when deploying this streamlit app locally: Streamlit didn't know what warehouse to use, despite this configuration here. The warehouse wasn't specified in my `config.toml` following the instructions in the existing `STREAMLIT.md`.

The solution was I configured `warehouse` in my `.snowflake/config.toml` file for the connection I'm using for the streamlt apps.

So, this is outside the scope of this PR, but necessary to update for local development: Add a warehouse configuration to your `config.toml` example in the `STREAMLIT.md`. Here is what my agent whipped up:

_Configure a `default` connection in `~/.snowflake/config.toml` with a role and warehouse that has access to the schemas the dashboard queries. The required role depends on the app — check the app's `snowflake.yml` for the `query_warehouse` and the schemas it reads from to determine the right role and warehouse. See the [[Snowflake CLI docs on configuring connections](https://docs.snowflake.com/en/developer-guide/snowflake-cli/connecting/configure-connections)](https://docs.snowflake.com/en/developer-guide/snowflake-cli/connecting/configure-connections) for full details._

```toml
[connections.default]
account = "<account>"
user = "<user>@sagebase.org"
authenticator = "PROGRAMMATIC_ACCESS_TOKEN"
token = "<your-programmatic-access-token>"
role = "<role>"  # e.g. DATA_ENGINEER for dashboards querying SAGE.GOVERNANCE
warehouse = "<warehouse>"  # e.g. STREAMLIT_XSMALL; check the app's snowflake.yml
```

philerooski: `snowflake.yml` is only referenced when deploying to Snowflake, not local deployments. Generally speaking, to use the Snowflake CLI tools (`snow`/`streamlit`) you'll need to set your `warehouse` in either the `config.toml` or `connections.toml` file. I'm not sure if it falls back to your user's default warehouse... but if it does you could probably skip updating your user config. Have you configured your default warehouse and if so, are you able to execute `snow` commands without setting the `warehouse` in `config.toml`? If so, then maybe this is a Streamlit specific set up step, rather than a more generic "how to configure your Snowflake connection" issue.

Technically, it doesn't matter which warehouse you use. But if you want to match the warehouse which the schema admin will use to execute the queries than you'll want to use `STREAMLIT_XSMALL`.

sage/elite/streamlit/elite_usage_metrics/environment.yml
```diff
@@ -0,0 +1,6 @@
+name: elite_usage_metrics
+dependencies:
+  - python=3.11.*
+  - snowflake-snowpark-python
+  - streamlit=1.*
+  - pandas
```
jaymedina, line 6: tldr; add `pyarrow` as a dependency so mamba handles it instead of `pip`

---

I was running into this issue where the streamlit app couldn't fetch pandas DataFrames from Snowpark:

Image

After a bit of fiddling with Claude, it seems like the issue was that `snowflake-snowpark-python` pulls in `pyarrow` as a dependency using pip, but pip's `pyarrow` was missing the native shared library `libutf8proc.2.dylib` on my Mac. The fix was to explicitly list `pyarrow` as a dependency so mamba miniforge (or miniconda which is what I used) manages the native library installation instead of pip. With that change, I was able to launch the streamlit app locally and load the widgets without issue:

Image

jaymedina (approved):
Dashboard looks as expected! Just had some comments related to setting a machine up to deploy this app locally.

sage/elite/streamlit/elite_usage_metrics/environment.yml
philerooski: Hmm, I couldn't reproduce in a `mambaorg/micromamba` container:
```
# run this locally from the repo root (assuming you have Snowflake credentials in connections.toml)
docker run --rm -it \
  -p 8501:8501 \
  -v ./sage/elite/streamlit/elite_usage_metrics:/home/mambauser/elite_usage_metrics \
  -v ~/.snowflake/connections.toml:/home/mambauser/.snowflake/connections.toml \
  mambaorg/micromamba bash

# then run this in the container and navigate to http://localhost:8501/ in your browser
micromamba create -f /home/mambauser/elite_usage_metrics/environment.yml -y
  micromamba run -n elite_usage_metrics_2 \
    streamlit run /home/mambauser/elite_usage_metrics/streamlit_app.py -- --local-dev
```

`snowflake-snowpark-python`, `pyarrow`, and `libutf8proc` are all available on conda-forge, so I don't understand why pip would be invoked or why you would need to install a system dependency (`libutf8proc` is bundled with the `libutf8proc` conda-forge package). It seems more like a micromamba configuration issue specific to your setup, which you were able to workaround by adding pyarrow as a dependency.

Perhaps try reinstalling micromamba or `micromamba clean --all` then recreate the environment?

sage/elite/streamlit/elite_usage_metrics/snowflake.yml
jaymedina: > are you able to execute snow commands without setting the warehouse in config.toml?

Yep! I was able to execute a `snow` SQL command with a config profile I have that doesn't configure a warehouse. 

I've removed the `warehouse` config from my default profile (which is what the streamlit app uses to connect to Snowflake), and I get the "_you don't have a warehouse configured_" message again, so yes it does look like it's a Streamlit-specific thing. I'd recommend updating the `config.toml` example in the `STREAMLIT.md` as part of this PR.

image

sage/elite/streamlit/elite_usage_metrics/environment.yml
jaymedina: I guess the issue comes from me using miniconda initially (conda's default channel is not conda-forge). Since the installation instructions are specific to the mamba context, and you were unable to reproduce this, I'll mark this as resolved.

Thanks for checking!

sage/elite/streamlit/elite_usage_metrics/streamlit_app.py
```diff
@@ -0,0 +1,1617 @@
...
+where
+    project_id in (
+        27229419,
+        52072575,
+        52072939,
+        52237024,
+        52642213,
+        53124793
+    )
+;
```
thomasyu888, line 75: Nit: easy optimization - can we only maintain this list one time somewhere in this streamlit app?

sage/elite/streamlit/elite_usage_metrics/streamlit_app.py
```diff
@@ -0,0 +1,1617 @@
...
+                    chart_df[value_col], errors="coerce"
+                )
+                chart_df = chart_df.dropna(subset=[value_col]).reset_index(drop=True)
+                chart_df["row_number"] = chart_df.index + 1
+
+                st.scatter_chart(
+                    chart_df.set_index("row_number")[[value_col]],
+                    width="stretch",
+                    height=400,
+                )
```
thomasyu888, line 1250: This doesn't really show anything informative

Image

sage/elite/streamlit/elite_usage_metrics/streamlit_app.py
```diff
@@ -0,0 +1,1617 @@
...
+@st.fragment
+def cell_8_2():
+    with st.container(border=True):
+        with st.container(
+            horizontal=True,
+            horizontal_alignment="distribute",
+            vertical_alignment="center",
+        ):
+            with st.container(height=80, border=False, vertical_alignment="center"):
+                st.markdown("### 2024-11-25 10:42am")
```
thomasyu888, line 1358: May want to remove this if we don't know what it is showing

thomasyu888 (approved):
🔥  Great work here, just minor comments (we can create JIRA tickets to resolve those comments if needed). We can also showcase the Streamlit app to Milan, just so she knows where to locate the app, and add more of her thoughts.

Should we add a disclaimer, as we did with the governance access submission dashboard, about modifying this in the UI?

sage/elite/streamlit/elite_usage_metrics/streamlit_app.py
philerooski: That's correct. It doesn't show anything informative in the "legacy" dashboard either.

The goal in transitioning the dashboards is to get them deployed as Streamlit apps in the same state as they are in the legacy dashboard. Cleaning up or fixing charts is beyond the scope of the more urgent issue: dashboards will disappear entirely in a few weeks.

sage/elite/streamlit/elite_usage_metrics/snowflake.yml
philerooski: Thanks for iterating on this with me. I realized it was just a one line change in STREAMLIT.md 🙃

PR #339, opened 2026-05-27
[SNOW-452] Transition Synapse Performance Metrics from dashboard to Streamlit

philerooski:
# Problem

Ticket: [SNOW-452](https://sagebionetworks.jira.com/browse/SNOW-452)

The Synapse Performance Metrics dashboard needed to be converted into a warehouse-runtime Streamlit app and integrated into the existing deployment flow.

# Solution

I converted the dashboard to a Streamlit app and addressed migration/runtime issues as part of that conversion.

Acceptance criteria coverage:
- Fetched and converted the app into `sage/product/streamlit/synapse_performance_metrics`.
- Switched Streamlit runtime compatibility to warehouse artifacts (`environment.yml`, `.streamlit/config.toml`, `snowflake.yml` updates).
- Replaced app SQL table references with `SYNAPSE_DATA_WAREHOUSE.SYNAPSE_EVENT.ACCESS_EVENT`.
- Added CI deploy matrix entry for the app in `.github/workflows/ci.yaml`.
- Added grant in `admin/grants.sql` so `SAGE_PRODUCT_ANALYST` has `USAGE` on the Streamlit object.

# Testing

You can test the Streamlit app by either examining it in Snowflake or deploying it locally.

The app has been deployed to Snowflake and is accessible by the schema's admin role.

You can also run it locally using the Streamlit local workflow described in `STREAMLIT.md`.

[SNOW-452]: https://sagebionetworks.jira.com/browse/SNOW-452


thomasyu888 (approved):
🔥 I'm approving and could we add a disclaimer for all of these apps to be a bit more explicit about the available management models?

Instead of only stating that manual changes will be overwritten, I'd suggest outlining two options:

1. Fully managed (recommended): changes are made via PRs to this repo, and any manual UI changes should be propagated back as code.  (For this particular dashboard, I'd opt to have platform create PRs, for others, we could create the PRs for them via a service desk ticket)
2. Unmanaged: CI is disabled, and owners can make direct changes, but they lose version history, code review, auditability, and other benefits of source-controlled configuration. (Create a DPE service desk ticket to have a fully unmanaged)

This gives dashboard owners a clear choice and sets expectations around the tradeoffs. Perhaps this is already included in the streamlit/readme.md

I'm was able to load the dashboard and it looks great!

philerooski:
@thomasyu888 

We could move this verbiage into the STREAMLIT.md file (would be significantly easier than maintaining it on a per-app basis), but I think we should consider more carefully whether we should support unmanaged apps as readily as we did for the legacy dashboard product, because it's apples and oranges.

With the legacy dashboard product, managed and unmanaged meant practically the same thing: either DPE was making the notebook-like edits or the analyst was making those same edits. The SQL code and access controls could only be managed in Snowsight and there was no cost for simply viewing the dashboard, only for running its queries.

With Streamlit apps, an unmanaged app has the same Snowsight-only downsides as the legacy dashboards, plus a few additional downsides:

1. Viewing the app consumes warehouse credits
2. The dashboard is no longer a point-and-click interface, it's a full-blown Python application. This makes the experience of editing the dashboard nearly identical to developing in the IDE, except the user is constrained to working in Snowflake Workspaces.

The one perk I can think of with the unmanaged approach is that users don't have to set up a local conda environment for their app. But given that we are all developing with agents now, this is not as much as a friction point as it used to be.

Personally, I think we should actively discourage unmanaged dashboards. If analysts are having difficulty developing upon their own dashboards (by cloning the repo and developing in their IDE), I think there are a few adjustments we can make before we might need to allow users to just edit their Streamlit directly in Snowsight.

jaymedina (approved):
This is neat! I did not know we had this dashboard.

I don't have access to the legacy dashboard to cross-reference, but the streamlit app loads fine and I don't see any missing content / error messages.

One quick nit: it's missing the little disclaimer that comes with DPE-managed streamlit apps at the top:

Image

Otherwise, approved

PR #346, opened 2026-06-02
[SNOW-438] Convert Audit Streamlit app to warehouse runtime

philerooski:
## Problem
The Governance Audit Streamlit app was not present in this repository as a warehouse-runtime deployable app, and related release configuration entries were missing.

## Solution
- Fetched and added the Audit app under `sage/governance/streamlit/audit`.
- Converted app artifacts to warehouse-runtime-compatible files:
  - `streamlit_app.py`
  - `environment.yml` (with `streamlit=1.*`)
  - `.streamlit/config.toml`
  - `snowflake.yml` with explicit identifier and `query_warehouse`.
- Updated SQL table references in the app to fully-qualified `synapse_data_warehouse` references where needed.
- Added required Streamlit usage grant for `SAGE.GOVERNANCE.AUDIT` in `admin/grants.sql`.
- Added deploy matrix entry for Audit app in `.github/workflows/ci.yaml` (`deploy_streamlit`).

## Testing
- Ran Snowflake object listing/describe to resolve and validate the app object.
- Verified runtime switched to `SYSTEM$WAREHOUSE_RUNTIME`.
- Validated required referenced tables exist and are non-empty.
- Ran app locally in headless mode and verified startup URL.
- Ran smoke log scan (`PASS: no error patterns matched`).


sage/audit/streamlit/audit/streamlit_app.py
thomasyu888, line 1: Image

Is it a separate ticket to manage the permissions of this app?

1. only GOVERNANCE + LT can access this
2. IP INFO -> PUBLIC

philerooski: I'll add grants for 1 but the changes for 2 are already here #348

philerooski: @thomasyu888 

Is it the case that LT ought to have access to all resources in `SAGE.GOVERNANCE` (in which case I can simply grant LT the SAGE_GOVERNANCE_ANALYST role)?

Or do we want to restrict LT to read privileges on certain objects only (in which case it's a little more complicated)?

thomasyu888: LT can have access to all resources in `SAGE.GOVERNANCE`.

philerooski: This (and a few other things) already happened in #311

thomasyu888: Thanks. The complicated part about this is that `TECH_PRODUCT` should not have access to this audit dashboard.

This almost goes back to asking what we should do about the current governance data access table in `SAGE.GOVERNANCE` and if it belongs there anymore.

thomasyu888 (approved):
🔥 LGTM, I validated the dashboard vs the streamlit app, and it looks great.

Quick question - is there a setting that makes these streamlit apps run each query in serial?

philerooski:
@thomasyu888 

I think you mean in parallel? And yes, you can configure that on a query-by-query basis. It's not always wise to parallelize everything blindly: we can take advantage of cached results from previous queries if applicable. I think in most cases we can parallelize, but I recommend we do it in a follow-up ticket.

thomasyu888:
> I think you mean in parallel? And yes, you can configure that on a query-by-query basis. It's not always wise to parallelize everything blindly: we can take advantage of cached results from previous queries if applicable. I think in most cases we can parallelize, but I recommend we do it in a follow-up ticket.

@philerooski yes, I meant parallel, and this makes sense.  From a product perspective, let's coordinate together on a "light product launch" notice once we're ready to "launch"

The notice can include something like: 

> For future iteration, please create tickets around
> 1. us managing modifying the streamlit app
> 2. requesting for features like parallel execution of SQL queries within your dashboard
> 3. Other Improvements?

jaymedina (approved):

sage/audit/streamlit/audit/streamlit_app.py
philerooski: @thomasyu888 

See my comment [here](https://sagebionetworks.jira.com/browse/SNOW-481?focusedCommentId=329405), but if we decide to keep this dashboard in the GOVERNANCE schema, I updated `admin/grants.sql` so that we revoke usage from SAGE_GOVERNANCE_ANALYST and instead directly grant usage to GOVERNANCE and SAGE_LEADERS.

philerooski:
Depends on #374 

Don't merge yet. We are going to move this to the AUDIT schema once #374 is merged.

PR #355, opened 2026-06-03
Merge dev into main

philerooski:
I want to merge dev into main for a few reasons:

1. I want to merge in the accumulation of behind the scenes changes we've made around agent files, CI, and dbt before we bring in a bulk of dashboard transition changes.
2. One of our dashboard transition PRs #337 is already in dev, so we'll get to test whether it can successfully be deployed. We only trigger the deployment step when merging to main, which is why we'll need to merge before we can officially say that our transition changes work.


thomasyu888 (approved):
🔥 LGTM!

PR #357, opened 2026-06-03
Grant rds access to product admin

philerooski:
# Problem

Our SAGE_PRODUCT_ADMIN role is responsible for executing Streamlit apps in the SAGE.PRODUCT schema, but some of these dashboards need read access on the RDS data.

# Solution

I grant database role `SYNAPSE_RDS_SNAPSHOT.raw_table_read_censored` to SAGE_PRODUCT_ADMIN. This is the same level of access which the TECH_PRODUCT role currently has.

# Testing

 
NA


thomasyu888 (approved):
🔥 LGTM!

PR #362, opened 2026-06-03
Grant SAGE.IT table privileges to SAGE_PRODUCT_ADMIN

philerooski:
# Problem

SAGE_PRODUCT_ADMIN needs read privileges on the tables in the SAGE.IT schema because we query those tables in the Synapse Sage Cost estimation dashboard.

# Solution

I granted read privileges on existing and future tables (via `GRANT ... ON ALL TABLES ...`).

# Testing


thomasyu888 (approved):
🔥 LGTM!

PR #377, opened 2026-06-17
[SNOW-510] Disable users

thomasyu888:
## Problem

People have left sage need to be disabled and gmails have been disabled
Ticket: https://sagebionetworks.jira.com/browse/SNOW-510

## solution
Disable users


philerooski (approved):
@thomasyu888 Are you manually disabling these users? If not, that will need to be captured in the SQL somewhere.

thomasyu888:
> @thomasyu888 Are you manually disabling these users? If not, that will need to be captured in the SQL somewhere.

@philerooski Ah I missed this... I disabled the users using the function you added?  I'll take a look in the UI

PR #385, opened 2026-07-01
[SNOW-505] Connect new `RDS_LANDING` tables to existing dbt models

jaymedina:
## Problem

Ticket: https://sagebionetworks.jira.com/browse/SNOW-505

The existing dbt models are built off `RDS_RAW` tables that have been manually created and populated by a locally run ingestion script that was used before setting up our automated ingestion pipeline.

Now that we have an established pipeline that's been running successfully and populating the landing tables in `RDS_LANDING` on-schedule, we need to modify our existing Snowflake infra to connect these objects to the staging/intermediate/mart layer objects in our catalog of products.

---

## Solution

The data layering convention that dbt recommends is:

source -> staging -> intermediate -> mart

Based on prior team conversations, notably [this thread](https://github.com/Sage-Bionetworks/snowflake/pull/326#discussion_r3318493234), here's a breakdown of how our objects, schemas, and object types map to each layer established by the dbt convention:

| Category | Object type | Location | Managed by | dbt layer |
|---|---|---|---|---|
| landing | table (append-only) | `RDS_LANDING` | COPY INTO task graph | — (not technically a dbt source... more like the source of the source) |
| raw | table (current state) | `RDS_RAW` | 🤷‍♀️ 🤷 | source |
| stage | view | `RDS_RAW` | dbt | staging |
| int | view | `RDS_RAW` | dbt | intermediate |
| mart | dynamic table | `SYNAPSE(_EVENT/_AGGREGATE)`, `SAGE` marts | dbt | mart |

As you can see from the table above, the missing component has been the relationship between `RDS_RAW` and `RDS_LANDING` source tables, and how we would populate the `RDS_RAW` source objects once the `RDS_LANDING` tables started coming in, which we had not yet finalized, according to the thread.

Since the `RDS_RAW` tables are source tables (i.e. provided by an external ETL tool and **not** managed by dbt) I'm creating an initial set of dynamic tables that are straight `SELECT *` clones of the `RDS_LANDING` tables.

I chose dynamic tables because they are simpler to maintain than tasks: Snowflake handles refresh scheduling automatically, and there's no additional task graph to wire up. If we later decide to add a transformation between `RDS_LANDING` and `RDS_RAW`, for QA for example, we could consider tasks instead for better observability.

That said, I continue to lean toward making `RDS_LANDING` the dbt source directly, and leaving the `RDS_RAW` space for the staging and intermediate layers (staging could be our dedicated QA checkpoint), but I'm following your (@philerooski) suggestion here and keeping the `RDS_RAW` (dynamic) tables as the dbt source layer.

**A final note before we get to the actual SQL scripts**: There are some `admin/` scripts in here, yet the branch is pointing to `dev`. I plan to leave these in here so you can look through all the scripts together to get the big picture view of all the changes, but if we agree on moving forward with dynamic tables, I can cherry pick them out into a separate PR to push directly to `main` before this one gets merged.

- **`synapse_data_warehouse/rds_raw/V2.72.0__create_rds_raw_dynamic_tables.sql`** — Drops the 14 manually-created placeholder tables in `RDS_RAW` and replaces them with dynamic tables that `SELECT *` from their `RDS_LANDING` counterparts. No transformations are applied; `TARGET_LAG = '8 hours'`, `WAREHOUSE = COMPUTE_XSMALL`.
- **`admin/ownership_grants/V1.36.0__rds_raw_dynamic_tables_ownership.sql`** — Transfers ownership of all dynamic tables in `RDS_RAW` to `PROXY_ADMIN` for both prod and dev. Dynamic tables, like tasks, must be owned by an account role.
- **`admin/future_grants/V1.36.1__rds_raw_future_dynamic_tables.sql`** — Sets future ownership of dynamic tables in `RDS_RAW` to `PROXY_ADMIN` and grants `SELECT, MONITOR` on future dynamic tables to `RDS_RAW_TABLE_READ` for both prod and dev.
- **`.github/workflows/test_with_clone.yaml`** — Adds `RDS_RAW` dynamic table ownership transfers to the clone setup step (matching the existing pattern for `SYNAPSE`, `SYNAPSE_AGGREGATE`, `SYNAPSE_EVENT`).

With all that said, this PR introduces new dynamic tables such that the final mapping of our Snowflake infra to the dbt convention looks like this:

| Category | Object type | Location | Managed by | dbt layer |
|---|---|---|---|---|
| landing | table (append-only) | `RDS_LANDING` | COPY INTO task graph | — (not technically a dbt source... more like the source of the source) |
| raw | dynamic table | `RDS_RAW` | Snowflake dynamic tables | source |
| stage | view | `RDS_RAW` | dbt | staging |
| int | view | `RDS_RAW` | dbt | intermediate |
| mart | dynamic table | `SYNAPSE(_EVENT/_AGGREGATE)`, `SAGE` marts | dbt | mart |

---

## Testing

before merge into `dev`:
- [x] Schemachange dry-run passes on the zero-copy clone for `synapse_data_warehouse/` scripts
- [x] `SHOW DYNAMIC TABLES IN SCHEMA RDS_RAW` confirms all 14 tables exist and are owned by `PROXY_ADMIN`
- [x] Dynamic tables refresh and contain data matching `RDS_LANDING` counterparts
- [x] `STG` views are re-populated to reflect the new dynamic tables, with minor transforms
- [x] `INT` views are re-populated to reflect the new `STG` views
- [x] mart objects are re-populated to reflect the new `INT` views
- [x] ensure `RDS_LANDING` tables' `snapshot_date` column has been updated with timezone (UTC)

after merge into `main`:
- [ ] `DATA_ENGINEER` role can `SELECT` from the dynamic tables via `RDS_RAW_TABLE_READ`
- [ ] `DATA_ANALYST` **does not** have access to these dynamic tables


philerooski (commented):
This is looking good overall, with a few small tweaks:

1. The RDS_RAW data is meant to reflect the current state of Synapse, i.e., data from the most recent snapshot. We capture every snapshot's data in RDS_LANDING, but all our dbt models are predicated on the source tables containing only a single snapshot-worth of data, that respects table constraints like uniqueness of primary keys etc. So rather than `select *`ing from RDS_RAW, we ought to `select * ... where snapshot_date = current_date()` (or maybe `current_date() - interval '1 day'`, depending on how Xa is exporting this data and when we are ingesting it). This maintains just the most recent snapshot in RDS_RAW and takes advantage of the `snapshot_date` cluster key which we configured earlier in the RDS_LANDING tables.
2. The target lag of these tables ought to be 8 hours, rather than 1 hour.

It's true that how RDS_LANDING relates to RDS_RAW in practice is a little fuzzy at the moment. We actually never got around to detailing this component of the pipeline in the design docs, although we knew from the start that we would need a pre-promotion validation layer (RDS_LANDING) before we let data flow downstream from the source tables (RDS_RAW).

What does this validation look like? We've talked before about needing to validate both the load integrity (e.g., task failures) and data quality (does the data we loaded look correct?), but exactly which checks we ought to implement and exactly how that should be implemented is To Be Designed. For the moment, I think it's reasonable that we can unblock ourselves by _assuming_ that the data quality is good and merely doing a light filtering of the most recent snapshot data between RDS_LANDING -> RDS_RAW.

jaymedina:
Thanks @philerooski - a few comments:

> The RDS_RAW data is meant to reflect the current state of Synapse, i.e., data from the most recent snapshot.

Just a heads up: the RDS snapshots also contain **immutable event-type data** (i.e. upload events, see [`MULTIPART_UPLOAD`](https://app.snowflake.com/mqzfhld/vp00034/#/data/databases/SYNAPSE_RDS_SNAPSHOT/schemas/PROD_595/table/MULTIPART_UPLOAD/data-preview) table), not just **mutable object-type data**, so that muddies the line separating `RDS_LANDING` from `RDS_RAW`. 

For the case of `MULTIPART_UPLOAD`, the RDS_RAW version of that table should look identical to RDS_LANDING's.

> The target lag of these tables ought to be 8 hours, rather than 1 hour.

I chose the target lag to be 1 hour because of Tom's [requirement](https://sagebionetworks.slack.com/archives/C0AFX3PFAFP/p1781647636776599?thread_ts=1776696837.675639&cid=C0AFX3PFAFP) of having fresh data to work with by 9am ET (6am PT) every day.

Since the ingest onto `RDS_LANDING` runs at 3am ET (12am PT) and currently the task graph doesn't take more than ~3 seconds to complete, that gives us about ~6 hours of latency we can afford. I'm comfortable pushing the target lag to 5 hours. Does that work?

jaymedina:
> What does this validation look like? We've talked before about needing to validate both the load integrity (e.g., task failures) and data quality (does the data we loaded look correct?), but exactly which checks we ought to implement and exactly how that should be implemented is To Be Designed. For the moment, I think it's reasonable that we can unblock ourselves by assuming that the data quality is good and merely doing a light filtering of the most recent snapshot data between RDS_LANDING -> RDS_RAW.

That works for me. Let's move forward with `RDS_RAW` as the source layer for now. I'll add that when we revisit the validation step, we should also consider how we can simplify the pipeline so that validation happens against RDS_LANDING as the source directly, rather than needing both RDS_LANDING and RDS_RAW.

If it just makes more sense for validation to happen in that intermediate RDS_RAW layer, let's move forward with that.

Do you want me to create a ticket?

philerooski:
> the RDS snapshots also contain immutable event-type data (i.e. upload events, see [MULTIPART_UPLOAD](https://app.snowflake.com/mqzfhld/vp00034/#/data/databases/SYNAPSE_RDS_SNAPSHOT/schemas/PROD_595/table/MULTIPART_UPLOAD/data-preview) table), not just mutable object-type data, so that muddies the line separating RDS_LANDING from RDS_RAW.

I see what you mean. When I union `MULTIPART_UPLOAD` from stacks 588 and 595, there's no overlap in their primary key:
```
with multipart_uploads as (
  SELECT *, 'PROD_595' AS SOURCE_TABLE
  FROM SYNAPSE_RDS_SNAPSHOT.PROD_595.MULTIPART_UPLOAD
  UNION ALL
  SELECT *, 'PROD_588' AS SOURCE_TABLE
  FROM SYNAPSE_RDS_SNAPSHOT.PROD_588.MULTIPART_UPLOAD
)
select source_table, id, count(*) as n from multipart_uploads
group by id, SOURCE_TABLE
having n > 1;
```

Which brings a few additional questions to mind:

1. When is data dropped from this table? Is it only when we migrate stacks (e.g., 594 -> 595) or is data dropped more frequently than that? It doesn't look like we're loading `SYNAPSE_DATA_WAREHOUSE_DEV.RDS_LANDING.MULTIPART_UPLOAD` data from the stage on a recurring basis yet so I'm not sure whether we can answer this question at the moment.
2. Assuming that this data refreshes on a predictable cadence -- like a complete refresh for every stack migration -- can we use a different dynamic table DDL that's equivalent to the "assume data is good quality" behavior of mutable object-type data types like `ACCESS_CONTROL`?
3. If data doesn't refresh on a predictable cadence, can we do something more advanced (like a MERGE) or does this data change too unpredictably for a daily snapshot? It may be the case that data is dropped more frequently than we snapshot it.

> I'm comfortable pushing the target lag to 5 hours. Does that work?

That's fine.

> we should also consider how we can simplify the pipeline so that validation happens against RDS_LANDING as the source directly, rather than needing both RDS_LANDING and RDS_RAW.

Tom and I discussed potentially being able to use dbt-managed tables in RDS_RAW, which reference dbt source tables in RDS_LANDING. I think this would probably work, but it would still require us to:

- Orchestrate a separate `dbt test` and `dbt run` invocation after data is loaded in RDS_LANDING
- It also assumes that the sort of validation that we would like to perform on the RDS_LANDING data can actually be accomplished with `dbt test`. For example, can we do record load counts so that we know the load task didn't silently fail?
- There could be some additional complications using dynamic tables in the mart layer with this setup. We might need to deploy DDL changes to the dynamic tables before we refresh the data in RDS_RAW, depending on how `dbt run` refreshes these tables.

So while it would be nice to have everything encompassed in the dbt data model, I'm not sure whether it would be much simpler than a standalone RDS_LANDING -> RDS_RAW pipeline. More to explore 🤷 

> Do you want me to create a ticket?

What did you have in mind? We run `dbt test` somewhere with the RDS_RAW tables as the source tables?

jaymedina:
Updated based on priv. comm. with Phil

---

@philerooski great points! what I meant by this:

> Do you want me to create a ticket?

Was: "_Do you want me to create a ticket that explores what the **validation step** is going to look like & revisits the earlier parts of our architecture (RDS_LANDING, RDS_RAW, and what these two schemas hold)?_"

So this ticket would have us explore:
* How do we handle the **immutable event-type data** in our validation step?
* Is it possible to simplify the pipeline so that validation happens against RDS_LANDING as the source directly?

These are necessary, but large considerations that blow up the scope of this PR. I also don't see them as urgent, since our current `STG_*` dbt models don't pull from tables that contain **immutable event-type data** (I validated this by eye, so it's worth double-checking).

Moving these considerations to a separate story will unblock this PR such that the only actionable items left are:
* Change `target_lag` to 5 hours
* Select only the latest snapshots (MUST be within the last day, so we can propagate issues with the landing tables down to the mart layer, via `NULL` column values) to move forward to the `RDS_RAW` source dynamic tables

jaymedina:
For record-keeping:

Phil and I agreed to move forward with this PR covering a simple design for connecting the `RDS_LANDING` tables to the `RDS_RAW` dynamic tables (the source layer of our dbt architecture). This design involves:

* Converting the `RDS_RAW` source tables to dynamic tables downstream from the `RDS_LANDING` tables
* Target lag of 5 hours
* `RDS_RAW` source dynamic tables are populated with only the latest screenshots of our `RDS_LANDING` tables (no older than a day)

Remaining questions of our design will be covered in this [Jira ticket](https://sagebionetworks.jira.com/browse/SNOW-521?atlOrigin=eyJpIjoiODUyMzJjMjM3NzgwNGZlMTlhMTZiYzIwM2M2YWUwYjUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ&issueKey=SNOW-521&subProduct=jira-software)

synapse_data_warehouse/rds_raw/V2.73.0__create_rds_raw_dynamic_tables.sql
```diff
@@ -0,0 +1,189 @@
...
+-- access_approval
+CREATE OR REPLACE DYNAMIC TABLE {{database_name}}.RDS_RAW.access_approval --noqa: JJ01,PRS,TMP
+    TARGET_LAG = '5 hours'
+    WAREHOUSE = COMPUTE_XSMALL
+    COMMENT = 'Dynamic table sourcing all columns from RDS_LANDING.access_approval with no transformations applied. Serves as the dbt source table for the stg_synapse__access_approval staging model.'
+AS
+SELECT *
+FROM {{database_name}}.RDS_LANDING.access_approval --noqa: JJ01,PRS,TMP
+WHERE snapshot_date >= DATEADD(day, -1, CURRENT_DATE())
+QUALIFY snapshot_date = MAX(snapshot_date) OVER ();
```
philerooski: Let’s use a strict equality predicate `=` for `snapshot_date`, based on how snapshots are assigned in `RDS_LANDING`.

The current combination of:

```sql
WHERE snapshot_date >= DATEADD(day, -1, CURRENT_DATE())
QUALIFY snapshot_date = MAX(snapshot_date) OVER ()
```

selects today’s snapshot when it exists but silently falls back to yesterday’s snapshot when it does not. Because each `RDS_RAW` table evaluates this independently, a silent load failure in one data type could leave some tables on today’s snapshot and others on yesterday’s snapshot.

Using:

```sql
WHERE snapshot_date = <expected snapshot date>
```

would instead make a missing snapshot visible by producing no rows. It would also make the `QUALIFY` clause unnecessary.

Was the fallback to the latest available snapshot intentional? Your current implementation aligns with your most recent comment:

> RDS_RAW source dynamic tables are populated with only the latest snapshots of our RDS_LANDING tables (no older than a day)

But I thought we agreed to be more strict and only include today's snapshot.

jaymedina: I think we need to confirm what time zone the system uses for `CURRENT_DATE()` in dynamic tables, but right now the assumption is there are no snapshots in `CURRENT_DATE()` so `DATEADD(day, -1, CURRENT_DATE())` _is_ the `<expected snapshot date>`. Here is a breakdown of our schedule:

8pm ET / 5pm PT / 2am UTC: Snapshots taken
-- max buffer: 4 hours --
2am ET / 11pm PT / 6am UTC: Copy to DPE buckets
-- max buffer: 1 hour --
3am ET / 12am PT / 7am UTC: Snowflake ingest of snapshots into `RDS_LANDING`
-- max buffer: 5 hours (this is our `TARGET_LAG` period) --
8am ET / 5am PT / 12pm UTC: Dynamic tables re-populated in `RDS_RAW`
-- max buffer: 1 hour --
9am ET / 6am PT / 1pm UTC: Snowflake tables are populated for use

The best way to explain this is with a concrete example:

**July 22, 2026**
* 5pm PT -- snapshots are taken
* 11pm PT -- snapshots copied to DPE buckets

**July 23, 2026**
* 12am PT -- `RDS_LANDING` tables are populated
* 5am PT -- `RDS_RAW` dynamic tables are populated <-- at this point `CURRENT_DATE()` is the day after the snapshots are taken, so the expected snapshot date is `DATEADD(day, -1, CURRENT_DATE())`

jaymedina: note from priv.comm.: encode time zone of `current_date()`

philerooski (approved):
~Did you push the changes we just discussed?~ Proactively approving pending those changes + Xa's timezone alignment

EDIT: I see them now.

jaymedina:
This PR is contingent on whether or not the `snapshot_date` (which is derived from Platform team's pipeline) is taken from a UTC timestamp or PT timestamp.

If UTC, we should set the `WHERE` filter to include the current day's snapshots, not the prior day's snapshots

jaymedina:
@philerooski since the `snapshot_date` we're extracting from platform team is [in UTC](https://sagebionetworks.slack.com/archives/C0AFX3PFAFP/p1784757934811579?thread_ts=1784754877.055949&cid=C0AFX3PFAFP), the snapshots are taken on the same day that the source dynamic tables are created, as far as Snowflake is concerned: even though our account's timezone is set to PT, `current_date()` will yield the same day as the latest `snapshot_date`, as long as our ingestion pipeline runs in a timely manner.

I've updated the SQL to filter for `snapshot_date = current_date()` and introduced a new schemachange script which updates our `RDS_LANDING` tables documenting that `snapshot_date` is derived from a UTC timestamp.

I don't see a need for updating our system to use UTC right now since we're already documenting what timezone `snapshot_date` is derived from. I worry it'll break the rhythm of our existing dynamic tables, AWS glue pipeline ingestion, or something else unforeseen, if we do something like switch the time zones in our account, but lmk what you think.

philerooski:
@jaymedina 

> and introduced a new schemachange script which updates our RDS_LANDING tables documenting that snapshot_date is derived from a UTC timestamp.

Awesome 🙌 

> even though our account's timezone is set to PT, current_date() will yield the same day as the latest snapshot_date

This works. It's a little unusual that we would compare a PT timestamp (`CURRENT_DATE()`) to a UTC timestamp (`snapshot_date`), although in either case -- `UTC = PT` vs `UTC = UTC` -- it's just a coincidence / contextually correct that we expect the dates to match: we're very much dependent on the timing of both Xa's export and when a refresh happens in RDS_RAW to grab the right snapshot. PT gives us a little more leeway to get data moving and aligns better with our actual working hours, so it seems like the best choice given the circumstances. 

> I don't see a need for updating our system to use UTC right now

There's definitely no need to update this on a global level.

Since this PR is doing more "real" work than the other PRs I opened and are waiting on your review, what do you think of merging this PR after those other PRs, assuming we can get everything in by Friday? That way we can get the easier stuff in before we bring in these more serious changes. If we're not able to merge those other PRs by tomorrow, I'll just merge this PR in on a Friday afternoon to minimize our chances of breaking downstream models while analysts are using them.

Also -- can you check the dev/clone build?

jaymedina:
Tests look good after a rebase onto `dev` with the new dbt auto-deploy in `test_with_clone`! I'm going to merge this into `dev` now.

There are a few loose threads we need to validate, but it can only happen after we make a Snowflake release to prod. I have these written out in the **Testing** section, but namely they are:

for [#391](https://github.com/Sage-Bionetworks/snowflake/pull/391/changes): ensure RDS_RAW_TABLE_READ_MASKED has SELECT, REFERENCES grants on dynamic tables in RDS_RAW
for [#391](https://github.com/Sage-Bionetworks/snowflake/pull/391/changes): ensure DATA_ENGINEER can still see row values for masked columns
for [#391](https://github.com/Sage-Bionetworks/snowflake/pull/391/changes): ensure DATA_ANALYTICS cannot see row values for masked columns

And center around making sure the new dynamic tables have the PII masking policy applied to them correctly. But we can't validate this until we merge to prod since there are admin/ scripts folded into this logic that need to be executed in order to work.

That means these loose threads will have to be addressed as part of a final Validation on the ticket, and I'll point to this comment when I write those instructions.

philerooski:
@jaymedina Awesome!! Let's verify that things look okay in dev and I can make a new release on main/prod later this afternoon once things have quieted down in Snowflake.

> making sure the new dynamic tables have the PII masking policy applied to them correctly

I don't think any of the tables in this PR ought to have a masking policy applied to them. Is this something we need to be concerned about at this point?

jaymedina:
@philerooski yep that sounds good 👍 

You're right, I see that these are [the tables](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4703649843/Governance+Restricted+Tables) whose columns are being masked depending on the role. NVM!

PR #381, opened 2026-06-22
[SNOW-480] Bootstrap `RDS_LANDING` tasks RBAC

jaymedina:
_Note: This PR description was drafted with AI and reviewed by me_ ✍️ 

----

## Problem

The initial `RDS_LANDING` task graph (root, proxy, 15 COPY INTO tasks, and finalizer) introduced in SNOW-480 had incomplete RBAC coverage.

## Solution

Leveraged the new `CLAUDE.md`/`AGENTS.md` instructional scripts to bootstrap the full RBAC pattern for `RDS_LANDING` tasks following the established proxy-admin ownership model:

- database_roles/V2.71.6 — creates RDS_LANDING_TASK_READ database role, assigns
ownership to RDS_LANDING_ALL_ADMIN, and grants it to RDS_LANDING_ALL_DEVELOPER
- rds_landing/V2.71.7 — grants MONITOR on all existing RDS_LANDING tasks to
RDS_LANDING_TASK_READ
- admin/ownership_grants/V1.32.0 — transfers ownership of all existing RDS_LANDING tasks
to SYNAPSE_DATA_WAREHOUSE[_DEV]_PROXY_ADMIN (tasks require an account role owner, not a
database role)
- admin/future_grants/V1.32.1 — grants ownership of future tasks to PROXY_ADMIN and
MONITOR on future tasks to RDS_LANDING_TASK_READ in both prod and dev

**Out of scope**: updated AGENTS.md with an explicit schemachange rule: scripts that reference an
object must have a higher version number than the script that creates it.

## Testing

**Can be tested now with clone**:
- [x] Schemachange dry-run passes on the zero-copy clone for synapse_data_warehouse/
scripts
- [x] SHOW TASKS IN SCHEMA RDS_LANDING confirms tasks are owned by PROXY_ADMIN

image

**Cannot be tested until merge into `dev`/`main`**:
- [ ] A user with RDS_LANDING_TASK_READ can query task history but cannot execute or
modify tasks
- [ ] SHOW FUTURE GRANTS IN SCHEMA RDS_LANDING reflects the new ownership and MONITOR
grants

PR #388, opened 2026-07-09
[SNOW-480] Move RBAC grants to `admin/` and improve `synapse_data_warehouse/AGENTS.md` instructions

jaymedina:
## problem

Address comments raised [here](https://github.com/Sage-Bionetworks/snowflake/pull/381#pullrequestreview-4647075914), namely:
* Privilege management for future and existing objects should exist in `admin/` -- database/schema level scripts are dedicated to creating and managing the objects themselves, and not their privileges
* Pretty-printed comments use up un-necessary tokens, and we should avoid them

## solution

- [x] Migrate RBAC grants to their appropriate SQL under `admin/`
- [x] Remove DDL from script
- [x] Update `AGENTS.md` to reflect these changes for future agents to follow

## testing

N/A


synapse_data_warehouse/rds_landing/V2.71.7__update_tasks_rbac.sql
```diff
@@ -1,7 +1,3 @@
...
-
--- Grant MONITOR privilege on all existing RDS_LANDING tasks to the task read database role.
-grant MONITOR
-    on all tasks
-    in schema RDS_LANDING
-    to database role RDS_LANDING_TASK_READ;
\ No newline at end of file
+-- Grant MONITOR on all existing RDS_LANDING tasks to the task read database role.
+-- Moved to admin/grants.sql — object-level grants (non-ownership, non-future) belong there,
+-- not in schemachange scripts.
```
philerooski, line 3: schemachange has a quirk where it will raise an error if there are no SQL statements in a script. Just add a `SELECT 1;` statement at the end.

synapse_data_warehouse/AGENTS.md
```diff
@@ -85,8 +81,13 @@ ALTER TASK {{database_name}}.schema.task_name RESUME;
 - Use `user_task_managed_initial_warehouse_size` — avoids needing a separate warehouse.
 - CRON uses `America/Los_Angeles` timezone.
 
+## Comments in schemachange scripts
+
+Keep comments short (one line max). Avoid step-numbered banners and ASCII decorators — they inflate context for agents reading these files.
```
philerooski, line 86: Enforcing this is a bit nuanced. There's multiple things we could consider:

- Omitting superfluous formatting should apply more generally than just `synapse_data_warehouse/` scripts.
- Comment style is a personal thing, and it may be more appropriate to enforce this at the user-level (i.e., in `~/.claude/CLAUDE.md`.
- The comment in `synapse_data_warehouse/rds_landing/V2.71.7__update_tasks_rbac.sql` in this PR is useful, but violates the one-line rule.
- It seems likely that the agent is getting the idea to pretty print comments by looking at existing scripts where comments are pretty printed. So it's trying to be smart by adopting an unnamed, existing convention.

My slight preference would be to not try to enforce a specific comment style at the repo-level, and instead enforce at the user-level. So we could remove this line if you think that's an effective solution.

synapse_data_warehouse/AGENTS.md
```diff
@@ -85,8 +81,13 @@ ALTER TASK {{database_name}}.schema.task_name RESUME;
...
 
 ## Constraints
 
 See root `CLAUDE.md` for schemachange rules that apply across all directories (version numbers, `CHANGE_HISTORY`, repeatable scripts, ownership transfers).
 
+Schema subdirectories (e.g. `rds_landing/`, `synapse_raw/`) must contain only DDL that creates or modifies objects. RBAC belongs elsewhere: role creation in `database_roles/`, grants in `admin/` (see `admin/AGENTS.md` for the full bootstrap split across `ownership_grants/`, `future_grants/`, and `grants.sql`).
```
philerooski: This is another convention which is bigger than just `synapse_data_warehouse/`, and we probably should be outlining where grants belong in the repo-root `AGENTS.md`. Unfortunately we can't just make a blanket statement that all grants belong in the `admin/` directory, since we have the convention of granting ownership and inheritance of database roles in the `{database}/database_roles` directories, as well.

If that makes sense, you could find an appropriate spot in the repo-root `AGENTS.md` to put this information.

jaymedina: sure thing! LMK what you think of the new placement

synapse_data_warehouse/AGENTS.md
jaymedina: > Comment style is a personal thing

I don't think it should be - the comments are not just for our future selves, but all future contributors, and I think should follow some type of standard across the repository, otherwise we'll run into unwanted inconsistencies like the one raised here. More of us are relying on agents to populate the codebase with comments, and I think having some established guideline for [at least the first draft] of a comment can be helpful with development.

> it's trying to be smart by adopting an unnamed, existing convention.

yes, I think that's exactly what happened. the agent saw my previous contributions and followed this pattern, which I think is why we should have an established pattern in an AGENTS.md somewhere to circumvent previous contributions. I think it makes sense to have this in the root AGENTS.md, and should include the 2 points you've raised:
1. No pretty-printing that consumes unnecessary tokens
2. Avoid comments that are mostly redundant with the SQL itself/don't add any specific contextual information

LMK what you think!

philerooski: That's reasonable. I like to give AI affirmative guidance -- rather than a list of "don't do this" -- so that it can start off on the right path rather than course correcting for rule violations. Something like:

> Comments should be plain, concise, and should add context that is not already obvious from the code itself. Prefer comments that explain intent, assumptions, business logic, or non-obvious implementation details.

jaymedina: Makes sense - [Updated](https://github.com/Sage-Bionetworks/snowflake/pull/388/commits/cbef8f5cfc3bf0f58ccad8d9914b7d753464dfbf)!

philerooski (approved):

PR #389, opened 2026-07-15
[SNOW-503] Configure dbt and deploy models as part of Github workflows

philerooski:
# Problem

Ticket: [SNOW-503](https://sagebionetworks.jira.com/browse/SNOW-503)

# Solution

## Acceptance Criteria
1. Design ticket if necessary
    * The changes are simple enough not to require a design doc
2. auto deploy DBT changes when PR is merged without having to run dbt run manually
    * DBT deployment occurs within the dev, prod, and sage environments directly after we perform a schemachange deployment. We also include a dbt deploy step in the `test_with_clone` workflow.

## Overview
I introduce a new `configure-dbt` action which serves a similar purpose as the preexisting `configure-snowflake-cli` action, except for configuring dbt. We invoke this action in CI job runs whenever we deploy to an environment which contains dbt models. Once dbt is configured, it's straightforward to call `dbt run` with the appropriate `--selector` and `--target` arguments.

# Testing

 
I prefer to test these changes in `dev`, rather than first merging a separate test workflow which uses the same action before merging the complete change set.

[SNOW-503]: https://sagebionetworks.jira.com/browse/SNOW-503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/configure-dbt/action.yml
+++ b/.github/actions/configure-dbt/action.yml
@@ -35,7 +35,10 @@ runs:
 
     - name: Install dbt-snowflake
       shell: bash
-      run: uv tool install "dbt-snowflake==1.11.6"
+      # dbt-snowflake has no console-script entry points of its own (dbt-core
+      # provides the `dbt` executable), so it must be installed as a --with
+      # dependency of dbt-core rather than as the tool package itself.
+      run: uv tool install dbt-core --with "dbt-snowflake==1.11.6"
 
     - name: Write private key file and dbt profiles.yml
       shell: bash
--- a/.github/actions/configure-dbt/action.yml
+++ b/.github/actions/configure-dbt/action.yml
@@ -20,7 +20,7 @@ inputs:
     description: "Database dbt should deploy to"
     required: true
   TARGET_NAME:
-    description: "Name of the dbt target/output to generate. Must be exactly 'prod' for sage models (gated on target.name == 'prod') to be eligible to run"
+    description: "Name of the dbt target/output to generate."
     required: true
   WAREHOUSE:
     description: "Snowflake warehouse"
@@ -72,7 +72,7 @@ runs:
               role: "${SNOWFLAKE_ROLE}"
               warehouse: "${SNOWFLAKE_WAREHOUSE}"
               database: "${SNOWFLAKE_DATABASE}"
-              schema: DBT_CI
+              schema: RDS_RAW
               private_key_path: "${PRIVATE_KEY_FILE}"
               private_key_passphrase: "${SNOWFLAKE_PRIVATE_KEY_PASSPHRASE}"
               threads: 4
--- a/.github/actions/configure-dbt/action.yml
+++ b/.github/actions/configure-dbt/action.yml
@@ -71,6 +71,8 @@ runs:
               user: "${SNOWFLAKE_USER}"
               role: "${SNOWFLAKE_ROLE}"
               warehouse: "${SNOWFLAKE_WAREHOUSE}"
+              # dbt requires us to set database/schema in the profile
+              # They can be overridden at the project, directory, or model levels
               database: "${SNOWFLAKE_DATABASE}"
               schema: RDS_RAW
               private_key_path: "${PRIVATE_KEY_FILE}"
--- a/.github/actions/configure-dbt/action.yml
+++ b/.github/actions/configure-dbt/action.yml
@@ -1,0 +1,82 @@
+name: "Configure dbt"
+description: "Installs dbt-snowflake and writes a ~/.dbt/profiles.yml for the transform project"
+inputs:
+  PRIVATE_KEY_PASSPHRASE:
+    description: "The pass phrase used to decrypt the private key"
+    required: true
+  PRIVATE_KEY:
+    description: "The private key"
+    required: true
+  ACCOUNT:
+    description: "Snowflake account identifier"
+    required: true
+  USER:
+    description: "Snowflake user identifier"
+    required: true
+  ROLE:
+    description: "Snowflake role dbt should assume (must match the deployment context, e.g. SYNAPSE_DATA_WAREHOUSE_ADMIN, SYNAPSE_DATA_WAREHOUSE_DEV_ADMIN, SAGE_ADMIN, or DATA_ENGINEER)"
+    required: true
+  DATABASE:
+    description: "Database dbt should deploy to"
+    required: true
+  TARGET_NAME:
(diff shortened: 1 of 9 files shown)


.github/actions/configure-dbt/action.yml
```diff
@@ -0,0 +1,82 @@
...
+              account: "${SNOWFLAKE_ACCOUNT}"
+              user: "${SNOWFLAKE_USER}"
+              role: "${SNOWFLAKE_ROLE}"
+              warehouse: "${SNOWFLAKE_WAREHOUSE}"
+              database: "${SNOWFLAKE_DATABASE}"
+              schema: RDS_RAW
+              private_key_path: "${PRIVATE_KEY_FILE}"
+              private_key_passphrase: "${SNOWFLAKE_PRIVATE_KEY_PASSPHRASE}"
+              threads: 4
+        EOF
```
thomasyu888, line 82: Just to be sure, does this print anything in the console? We don't want it actually printing the private key passphrase

.github/actions/configure-dbt/action.yml
```diff
@@ -0,0 +1,82 @@
...
+          target: ${DBT_TARGET_NAME}
+          outputs:
+            ${DBT_TARGET_NAME}:
+              type: snowflake
+              account: "${SNOWFLAKE_ACCOUNT}"
+              user: "${SNOWFLAKE_USER}"
+              role: "${SNOWFLAKE_ROLE}"
+              warehouse: "${SNOWFLAKE_WAREHOUSE}"
+              database: "${SNOWFLAKE_DATABASE}"
+              schema: RDS_RAW
```
jaymedina, line 78: nit: add comment explaining why this needs to be hardcoded, and explaining it's an arbitrary schema

thomasyu888 (commented):
🔥 Thanks for doing this, and I just had one comment. I'm excited about this change

.github/actions/configure-dbt/action.yml
philerooski: It does not leak any sensitive data, for two reasons:

1. The passphrase is only ever written into the `profiles.yml` file, never printed to the console with echo or similar, and the script doesn't have debug/verbose logging turned on that could leak it.
2. It comes from `secrets.ADMIN_SERVICE_PASS_PHRASE` in the workflow. GitHub Actions automatically replaces any log output that matches a secret's value with ***, so even if it were printed by accident, it would show up masked, not in plain text.
3. Similar reasoning for the private key.

We also restrict permissions on `profiles.yml` (chmod 0600) right after writing it, so only the file's owner can read it — no other user or process on the runner can access it.

philerooski:
Cleaned up the shared `configure-dbt` / `configure-snowflake-cli` composite actions for consistency:

- **`configure-dbt`**:
    * switched to `uv tool install` (matching `configure-snowflake-cli`'s tooling) instead of `setup-python` + `pip install`
    * added a `dbt debug --target <target>` step so a bad role/account/warehouse fails fast right after configuration instead of only surfacing later in `dbt run`.
- **`configure-snowflake-cli`**:
    * now `chmod 0600`s the private key file after writing it (it was the only one of the two actions leaving the key world-readable)
    * dropped the unused `outputs:` block — nothing in `ci.yaml`/`test_with_clone.yaml` referenced them, and the same info (paths, version, connection test) was already visible in the step logs.

philerooski:
@jaymedina You can see a successful run of the `test_with_clone` workflow with these changes [here](https://github.com/Sage-Bionetworks/snowflake/actions/runs/30050571449/job/89351556462).

.github/workflows/ci.yaml
```diff
@@ -81,8 +97,25 @@ jobs:
...
             --snowflake-role synapse_data_warehouse_admin
 
+      - name: Configure dbt
+        uses: ./.github/actions/configure-dbt
+        with:
+          PRIVATE_KEY_PASSPHRASE: ${{ secrets.ADMIN_SERVICE_PASS_PHRASE }}
+          PRIVATE_KEY: ${{ secrets.ADMIN_SERVICE_PRIVATE_KEY }}
+          ACCOUNT: ${{ secrets.SNOWSQL_ACCOUNT }}
+          USER: ${{ vars.ADMIN_SERVICE_USER }}
+          ROLE: synapse_data_warehouse_admin
```
jaymedina: nit: why is this hard-coded? do we want to use `${{ vars.SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE }}_admin` ?

jaymedina (approved):
lgtm! just minor comments, and one doc suggestion:

I think it would be helpful to mention in the user/agent docs when you to use `schemachange` V scripts, and when to use the models in `transform/` -- I could see the agent or a new contributor getting confused and working with `schemachange` while contributing a product to the `INT` layer, for example. can we make that distinction in the docs more explicit?

.github/workflows/ci.yaml
philerooski: These arguments are specific to deploying dbt within the `schemachange_synapse_data_warehouse_prod` job, but since we define the same value at the workflow level it's a good idea to reference it here. I'll update.

philerooski:
@jaymedina 

Good idea. I think this works best in the repo-root AGENTS.md, since whether an agent starts working in `synapse_data_warehouse/` or `sage/` vs. working in `transform/` is going to immediately differentiate whether we ought to be working with the schemachange or dbt framework.

I reviewed this AGENTS.md and I thought that, overall, there was already a pretty good characterization of the relationship between schemachange and dbt. I thought this content plus the additional contextual information an agent can get from comparing a user's prompt to how things are already implemented gives the agent enough information to make an informed decision about when to use schemachange vs. dbt. I asked Claude to explain the relationship based solely on the repo-root AGENTS.MD and was pretty satisfied with its response. I did make one small update for clarity.

But then I started questioning how we could measure whether our AGENTS.md files actually affect agent behavior when performing an implementation, beyond just "the concept seems understandable based on existing documentation". So I went a little overboard and had Claude administer an exam to independent agents. I provided three "exam prompts" or plausible real-world implementation requests, then provided my (human/expert) reasoning around why it might make more sense to implement the request within the schemachange vs. dbt framework. Then I had Claude synthesize how well the agents did and worked with it to update AGENTS.md with any gaps.

I felt like it worked pretty well! But just to be sure, I had Claude re-administer the exams, now with the updated documentation, to see whether our changes made a positive difference. Overall, they did guide agents to reason more like a human/expert, and I even turned up an additional gap in one of our AGENTS.md files, which I patched.

I want to talk more about this process in Wednesday's Code Space's meeting if you're interested in hearing more details.

PR #390, opened 2026-07-16
[SNOW-482] Create RDS_RAW PII masking policies

philerooski:
# Problem

Ticket: [SNOW-482](https://sagebionetworks.jira.com/browse/SNOW-482)

We want to grant `TECH_PRODUCT`/`SAGE_LEADERS` read access on tables in `RDS_RAW` but they shouldn't be able to see sensitive data in [these columns or tables](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4703649843/Governance+Restricted+Tables).

None of these 9 tables with sensitive columns exist in `RDS_RAW` yet — this PR only creates the masking-policy infrastructure so it's ready whenever those tables land.

# Solution
Adds `admin/policies/V1.36.0__rds_raw_pii_masking_policies.sql`:
1. **Create masking policies**: `PII_MASK_TEXT`/`PII_MASK_BINARY` masking policies in both `SYNAPSE_DATA_WAREHOUSE.RDS_RAW` and `SYNAPSE_DATA_WAREHOUSE_DEV.RDS_RAW`: the real value is returned only when the session holds the existing `RDS_RAW_TABLE_READ` database role (full/uncensored access); every other role, including the new `RDS_RAW_TABLE_READ_MASKED` role (see the companion PR #391 targeting `dev`), sees `NULL`.
2. **Grant the schema admin ownership privileges on these policies**: (`admin/ownership_grants/V1.36.1__rds_raw_pii_masking_policies_ownership.sql`):
    * As usual, we give the schema admin ownership over its objects. In a future deployment, we can attach these policies to sensitive columns once those tables are added to `RDS_RAW`.

No column attachment happens in this PR — there's no tables to attach to yet.

See the companion PR #391 for follow-up steps once we have the masking policies established.

# Testing
After merge, verify in a Snowflake worksheet:
```sql
DESC MASKING POLICY SYNAPSE_DATA_WAREHOUSE.RDS_RAW.PII_MASK_TEXT;
DESC MASKING POLICY SYNAPSE_DATA_WAREHOUSE.RDS_RAW.PII_MASK_BINARY;
SHOW GRANTS ON MASKING POLICY SYNAPSE_DATA_WAREHOUSE.RDS_RAW.PII_MASK_TEXT;
-- expect OWNERSHIP granted to DATABASE ROLE RDS_RAW_ALL_ADMIN
```
Repeat for `SYNAPSE_DATA_WAREHOUSE_DEV`.

[SNOW-482]: https://sagebionetworks.jira.com/browse/SNOW-482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


admin/policies/V1.36.0__rds_raw_pii_masking_policies.sql
```diff
@@ -0,0 +1,33 @@
...
+--
+-- The real value is returned only when the querying
+-- session holds the database role RDS_RAW_TABLE_READ (the uncensored read role);
+-- otherwise all values are returned as NULL.
+
+CREATE MASKING POLICY IF NOT EXISTS SYNAPSE_DATA_WAREHOUSE.RDS_RAW.PII_MASK_TEXT
+    AS (val VARCHAR) RETURNS VARCHAR ->
+    CASE
+        WHEN IS_DATABASE_ROLE_IN_SESSION('RDS_RAW_TABLE_READ') THEN val
+        ELSE NULL
```
thomasyu888, line 11: Personally, I'm not a huge fan of NULL because then we don't know if it's a blank or a censored value.  Thoughts on this?

thomasyu888 (commented):
🔥 LGTM, I'll defer to @jaymedina for final review of this, and I appreciate the work to implement the masking policies.

Note: the overall risk is lowered by the fact that platform does this: any tables containing ‘secrets’ should be either encrypted or be salted hashes (but this needs to be confirmed)

admin/policies/V1.36.0__rds_raw_pii_masking_policies.sql
philerooski: I have a slight personal preference for NULL. Here's why:

I think NULL makes it explicit that this data point isn't referencable. There's sort of an equal ambiguity with setting these values to something like `***` because then it's unclear whether those values are actually set that way in the backend or whether this is the result of some transformation. We aren't able to grant read-only privileges on masking policies, so analysts wouldn't be able to figure this out on their own.

It's only the analysts who will see NULL, so there won't be any ambiguity for developers.

To alleviate any potential confusion, we could add a projection policy which hides the column entirely, but then there's the additional overhead of managing projection policy objects and possible downstream effects since the table schema looks different depending on who's viewing it.

My preference is to wait and see whether there's any confusion around these masked values before we start implementing solutions to that confusion. They are pretty esoteric data points to begin with, so we'll see if anybody actually wants to reference them.

thomasyu888: Thanks for your explanation, it is very helpful, and I'm aligned now.

admin/policies/V1.36.0__rds_raw_pii_masking_policies.sql
```diff
@@ -0,0 +1,33 @@
+-- PII masking policies for the RDS_RAW schema
+--
+-- The real value is returned only when the querying
+-- session holds the database role RDS_RAW_TABLE_READ (the uncensored read role);
+-- otherwise all values are returned as NULL.
+
+CREATE MASKING POLICY IF NOT EXISTS SYNAPSE_DATA_WAREHOUSE.RDS_RAW.PII_MASK_TEXT
+    AS (val VARCHAR) RETURNS VARCHAR ->
+    CASE
+        WHEN IS_DATABASE_ROLE_IN_SESSION('RDS_RAW_TABLE_READ') THEN val
```
jaymedina, line 10: does this account for secondary roles?

philerooski: [Yes](https://docs.snowflake.com/en/sql-reference/functions/is_database_role_in_session).

> Verifies whether the database role is in the user’s active primary or secondary role hierarchy for the current session...

jaymedina (approved):
lgtm pending concern about how secondary roles impact the mask now that we've enabled those

PR #396, opened 2026-07-23
Revoke unnecessary roles from DPE_SERVICE

philerooski:
# Problem

This user is granted SYSADMIN and SECURITYADMIN roles, but only ever uses DATA_ANALYTICS and DATA_ENGINEER roles to execute its queries (presumably from Looker Studio).

# Solution

I revoked these unnecessary roles.

# Testing

NA

### DIFF
--- a/admin/grants.sql
+++ b/admin/grants.sql
@@ -15,8 +15,13 @@ TO USER "x.schildwachter@sagebase.org";
 GRANT ROLE SYSADMIN
 TO USER "phil.snyder@sagebase.org";
 
-GRANT ROLE SYSADMIN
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 thanks for this + I'm assuming you looked within the snowflake usage telemetry to determine that these roles arent used by these users?

philerooski:
@thomasyu888 

> I'm assuming you looked within the snowflake usage telemetry to determine that these roles arent used by these users?

That's right. I found that in the past 2 months the only roles used were DATA_ANALYTICS (for ~90%) of the queries and DATA_ENGINEER.

PR #397, opened 2026-07-28
[SNOW-505] Grant SELECT, MONITOR on current `RDS_RAW` dynamic tables

jaymedina:
# problem

Current RBAC bootstrapping for `RDS_RAW` dynamic tables is limited to `admin/future_grants/` scripts (#385). However, this script was pushed to `dev` alongside another script that will create the dynamic tables before the future grants can be applied, making them void for these tables. 

To fix this, we need to apply this grant not just for future dynamic tables, but also current dynamic tables, which is what our dynamic tables will be by the time `admin/grants.sql` runs.

# solution

See proposed solution here: https://github.com/Sage-Bionetworks/snowflake/pull/397#issuecomment-5118788779

# testing

N/A

PR #398, opened 2026-07-30
[SNOW-505] Remove `RDS_RAW_TABLE_READ` grants & bump up V on `RDS_RAW_TABLE_READ_MASKED` grants

jaymedina:
## problem

This is a follow-up PR for #397, and addresses the following issues:
* `RDS_RAW_TABLE_READ` grants to dynamic tables in the schema are instead applied directly to `main/` following conversation in #397 so these SQL scripts must be removed from `dev/`
* `RDS_RAW_TABLE_READ_MASKED` grants to dynamic tables in the schema will remain in `dev/` and therefore will run after the grants from the previous point, meaning their schemachange V needs to be bumped up

## solution

- [x] Remove the `admin/` scripts from #397 that still exist in `dev`
- [x] Bump up `RDS_RAW_TABLE_READ_MASKED` grants SQL script to `V1.38.X`
- [x] Remove backfill grants on `RDS_RAW` tables in `grants.sql` ... we no longer have tables in `RDS_RAW`

## testing

N/A since the changes made here are on `admin/` scripts that don't get deployed against our `dev` environment


philerooski (approved):
One small change: let's remove the grant on future tables from `admin/future_grants/V1.38.0__rds_raw_future_tables_masked_read.sql`. (The future grants on dynamic tables are fine).

PR #403, opened 2026-07-31
[SNOW-544] Fix dbt database resolution and dynamic table refresh boundary

philerooski:
# Problem
The `schemachange_sage` CI job's `Run dbt models` step fails on `main`:

```
Database Error in model data_access_submission_dashboard (models/marts/sage/data_access_submission_dashboard.sql)
002003 (02000): SQL compilation error:
Schema 'SAGE.RDS_RAW' does not exist or not authorized.
```
Run: https://github.com/Sage-Bionetworks/snowflake/actions/runs/30661534943/job/91258878835

Two separate bugs cause this:
1. `staging.synapse`/`intermediate.synapse` models have no explicit `+database` override in `dbt_project.yml`, so they resolve against the connection's default database, which was set to `SAGE` in this case, so `ref('int_synapse_data_access_submission_enriched')` compiled to `SAGE.rds_raw...` instead of `SYNAPSE_DATA_WAREHOUSE.rds_raw...`. 
    + This bug has existed since the model was authored ([SNOW-407](https://sagebionetworks.jira.com/browse/SNOW-407)), but was masked because dbt only recreates an unchanged dynamic table on `--full-refresh` or a config change — which the recent full-refresh hotfix now forces.
2. The same bug we fixed in #400, except for the `data_access_submission_dashboard` model, which we materialize as a dynamic table in the `SAGE.GOVERNANCE` schema.

Ticket: [SNOW-544](https://sagebionetworks.jira.com/browse/SNOW-544)

# Solution
1. `data_access_submission_dashboard` refreshes without the `SAGE.RDS_RAW does not exist` error.
    * Updated the default database used in the dbt profile. Since our `marts.sage` models explicitly override `+database: SAGE`, we only need to specify the "default" database for our dbt models, which is a SYNAPSE_DATA_WAREHOUSE environment. In practice, this means the SYNAPSE_DATA_WAREHOUSE _database_, since we only deploy `marts.sage` models in production.
2. `data_access_submission_dashboard` refreshes without the "Dynamic Tables do not support reading from views containing objects of type DYNAMIC_TABLE" error.
    * Wrapped all four refs in [transform/models/marts/sage/data_access_submission_dashboard.sql](../blob/snow-544-fix-sage-dashboard-refresh/transform/models/marts/sage/data_access_submission_dashboard.sql) in `dynamic_table_refresh_boundary(...)`, matching the SNOW-542 precedent.

# Testing
- `dbt parse` and `dbt compile --select data_access_submission_dashboard` succeed locally with no new errors, and the compiled SQL correctly resolves refs and preserves the `dynamic_table_refresh_boundary(...)` wrap.
- This PR is opened as a draft; `test_with_clone.yaml` only exercises the `synapse_data_warehouse` selector against the clone, not `sage`, so full end-to-end verification requires merging this PR, then confirming `schemachange_sage` succeeds on the next `main` deploy.

[SNOW-407]: https://sagebionetworks.jira.com/browse/SNOW-407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SNOW-542]: https://sagebionetworks.jira.com/browse/SNOW-542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SNOW-544]: https://sagebionetworks.jira.com/browse/SNOW-544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -146,7 +146,11 @@ jobs:
           ACCOUNT: ${{ secrets.SNOWSQL_ACCOUNT }}
           USER: ${{ vars.ADMIN_SERVICE_USER }}
           ROLE: sage_admin
-          DATABASE: SAGE
+          # marts/sage models explicitly override +database: SAGE in
+          # dbt_project.yml, so the default database here only needs to
+          # resolve refs into staging/intermediate synapse models, which
+          # live in SYNAPSE_DATA_WAREHOUSE.
+          DATABASE: ${{ vars.SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE }}
           TARGET_NAME: prod
 
       - name: Run dbt models
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -146,10 +146,8 @@ jobs:
           ACCOUNT: ${{ secrets.SNOWSQL_ACCOUNT }}
           USER: ${{ vars.ADMIN_SERVICE_USER }}
           ROLE: sage_admin
-          # marts/sage models explicitly override +database: SAGE in
-          # dbt_project.yml, so the default database here only needs to
-          # resolve refs into staging/intermediate synapse models, which
-          # live in SYNAPSE_DATA_WAREHOUSE.
+          # Resolve refs for staging/intermediate models,
+          # which live in a SYNAPSE_DATA_WAREHOUSE environment.
           DATABASE: ${{ vars.SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE }}
           TARGET_NAME: prod
 
--- a/transform/models/marts/sage/data_access_submission_dashboard.sql
+++ b/transform/models/marts/sage/data_access_submission_dashboard.sql
@@ -17,32 +17,32 @@ WITH base AS (
         state_reason,
         accessor_changes,
         data_access_submission_raw
-    FROM {{ ref('int_synapse_data_access_submission_enriched') }}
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -146,7 +146,9 @@ jobs:
           ACCOUNT: ${{ secrets.SNOWSQL_ACCOUNT }}
           USER: ${{ vars.ADMIN_SERVICE_USER }}
           ROLE: sage_admin
-          DATABASE: SAGE
+          # Resolve refs for staging/intermediate models,
+          # which live in a SYNAPSE_DATA_WAREHOUSE environment.
+          DATABASE: ${{ vars.SNOWFLAKE_SYNAPSE_DATA_WAREHOUSE_DATABASE }}
```
philerooski, line 151:
```
          # marts/sage models explicitly override +database: SAGE in
          # dbt_project.yml, so the default database here only needs to
          # resolve refs into staging/intermediate synapse models, which
          # live in SYNAPSE_DATA_WAREHOUSE.
```

thomasyu888 (approved):
🔥 Talked this through in a live review. DBT requires the dbt_project to specify a database and it's an artifact of DBT even if we're running just the sage models.

Separately, the `--full-refresh` triggers all upstream dynamic tables, it will do `create or replace` on models that are changed and refresh the upstream tables.

